PR #22, opened 2015-01-08
Added argument sendTreeHasChanged=true to TreeViewItem::removeSubItem

jzrake:
It may be desirable to delay refreshing and repainting the TreeView. In
particular, deadlocks can be caused by repainting the TreeView component
into an OpenGL context while it is active on the render thread.

### DIFF
--- a/modules/juce_gui_basics/widgets/juce_TreeView.cpp
+++ b/modules/juce_gui_basics/widgets/juce_TreeView.cpp
@@ -1226,13 +1226,13 @@ void TreeViewItem::addSubItem (TreeViewItem* const newItem, const int insertPosi
     }
 }
 
-void TreeViewItem::removeSubItem (int index, bool deleteItem)
+void TreeViewItem::removeSubItem (int index, bool deleteItem, bool sendTreeHasChanged)
 {
     if (ownerView != nullptr)
     {
         const ScopedLock sl (ownerView->nodeAlterationLock);
 
-        if (removeSubItemFromList (index, deleteItem))
+        if (removeSubItemFromList (index, deleteItem) && sendTreeHasChanged)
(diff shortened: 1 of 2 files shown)


julianstorer:
None of the GUI classes are thread-safe! If you're calling it from a
different thread then there will be a million other ways in which the class
could also fail or deadlock, so that's not a valid justification for a
change like this!

It also doesn't seem to me like a sensible thing to just add a bool flag to
disable the callback. Perhaps it could be argued that the callbacks
could/should be made asynchronously rather than synchronous, but such a
change would probably break existing code.

Ro0mquy:
You're probably right, changing existing behaviour would be a bad thing. But I think adding a switch to the Value/ValueTree functions for turning off callbacks (like for example in the button class) would be a good thing. If done via optional argument they would not break existing code.

jzrake:
Hi Jules and thanks so much for the quick response -

The deadlock I am experiencing was not caused by modifying any GUI class instances (TreeViewItem's) on the render thread. It happens when (1) the message thread issues a TreeView::repaint via removeSubItem -> treeHasChanged(), and (2) the TreeView is being painted into an OpenGLRenderer component. I have checked that if the TreeView is painted into a non-OpenGLRenderer component, then the deadlock does not occur.

For the moment, I am convinced that the only way I can avoid it is to suppress the automatic repaint. Another work-around (without modifying the API) would be to make the TreeView momentarily invisible while its children are being modified, but this would require additional book-keeping.

I understand if this workaround should not be part of the API. Is there a way to suppress the repaint in any other way?

jzrake:
Oops, I'm sorry I just now saw post in the wiki about not accepting pull requests. If this needs further discussion we can move it to the forum.

PR #28, opened 2015-04-22
Fixed missing install paths on OS X.

tpoole:
The final step of the build process on OS X produces errors if the install destinations don't already exist. This creates directories in ~/Library/Audio/Plug-Ins for AU, VST and VST3, and is more careful about RTAS.


julianstorer:
Just noticed this pull request - sorry, didn't realise you had posted it! Thanks - it's a good idea, we'll add something along these lines!

PR #30, opened 2015-05-15
Add some missing override specifiers

vincent23:


julianstorer:
Thanks Vincent - we can't accept pull requests but will clean up these warnings ourselves!

vincent23:
Thank you for cleaning up the warnings. Do you have a specific reason for dropping the 'override' specifiers on getApplicationName() and getApplicationVersion(), as I'm getting a lot of annoying warnings from clang++ when compiling my project?

julianstorer:
Thanks Vincent - I was mainly worried in that case, in case older compilers
would complain about the edge-case of marking a pure virtual as an
override. I may be worrying about that unnecessarily though!

PR #31, opened 2015-06-28
[Introjucer] fixed warning C4510 on Visual Studio 2013

OMGtechy:
Also moved the #define for JUCE_DELETED_FUNCTION into MSVC >= 1800 block, since it's supported and compiles just fine.


julianstorer:
Cheers Joshua! Will clean those things up..

PR #32, opened 2015-06-30
Creation of already existing named piped.

noizefloor:
It was possible to create named pipes that already exist.

### DIFF
--- a/modules/juce_core/native/juce_win32_Files.cpp
+++ b/modules/juce_core/native/juce_win32_Files.cpp
@@ -795,9 +795,20 @@ public:
           connected (false), ownsPipe (createPipe), shouldStop (false)
     {
         if (createPipe)
+        {
             pipeH = CreateNamedPipe (filename.toWideCharPointer(),
                                      PIPE_ACCESS_DUPLEX | FILE_FLAG_OVERLAPPED, 0,
                                      PIPE_UNLIMITED_INSTANCES, 4096, 4096, 0, 0);
+            if(GetLastError() == ERROR_ALREADY_EXISTS)
+            {
+                disconnectPipe();
+                if (pipeH != INVALID_HANDLE_VALUE)
+                {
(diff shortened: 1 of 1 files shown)


julianstorer:
Thanks for the heads-up, will look into this!

noizefloor:
Hi Julian,

The for the fix. I'm currently looking at the named pipe implementation on
OS X and it has the same behavior. The creation of a pipe succeeds, even if
the pipe already exists. As pipes on Unix systems are just files this
behavior would make sense in case one of these files was left in the tmp
directory. Somehow I get the feeling that this behavior might be by design,
so it might be a good idea to add a Boolean parameter "mustNotExist" with a
default value of false to the createPipe method.

Cheers
Jan
Am 30.06.2015 16:07 schrieb "Julian Storer" <notifications@github.com>:

julianstorer:
Ok, thanks, will do some sanity-checking.

noizefloor:
You also might wanna look at the destructor of the native pipe
implementation for POSIX systems. On destruction "unlink" is called, even
if the fifos could not be created. Right now that doesn't really causes
problem, but it's going to cause some trouble if the creation of pipe
should fail, if the pipe already exists.

Example:

Process A tries to creates pipe "myPipe". The pipe doesn't exists, so the
creation succeeds.

Process B tries to create the pipe "myPipe". The pipe already exists,
creation fails, "Pimpl" gets destroyed and unlinks the files.

Process C tries to create the pipe "myPipe". As the files are unlinked, at
least File::exists will return false as it uses access (), even if the
Process A is still using them and they are not deleted yet. I think mkfifo
got the same behavior. Long story short, the creation succeeds, even if the
pipe already exists.

Cheers
Jan
Am 01.07.2015 11:21 schrieb "Julian Storer" <notifications@github.com>:

> Ok, thanks, will do some sanity-checking.
>

PR #33, opened 2015-07-01
Added the parameter 'mustNotExist' to InterprocessConnection:: createPipe

noizefloor:
This parameter is set to 'false' by default. If it's set 'true' the creation of the pipe fails if another pipe with the same name already exists.

This means the default behavior is the same as it was before fc2ce885767b932eb60ac67df506fd2e51611fe5 was checked in. "createPipe" creates a pipe if is not existing and connect to it if does exists.

If mustNotExist is set to true the method creates a pipe if the pie does not exist, otherwise it returns false.


noizefloor:
Hi Julian,

have you already had the time to look into this?

julianstorer:
Hi Jan

I started, but never finished! It's not forgotten, will do it soon!

julianstorer:
FYI, I've done something equivalent to this now - let me know if you have any problems with it!

noizefloor:
It seems to work on Windows, but I'm not able to test it under OS X as I cannot compile against a 10.6 SDK anymore, since the Apple CoreAudio support files where moved into the juce_audio_plugin. I was using an older version for 10.6.

julianstorer:
<sigh..> Yes, we heard about the 10.6 thing - will sort that out today.

PR #34, opened 2015-07-03
Update juce_PopupMenu.cpp

gvisol:
It seems that one parent level more is needed in order to reach dismissMenu method. It was tested on my project - Win32 & OSX x86 & x64

The Component parent hierachy is:
        this | [juce::PopupMenu::HelperClasses::NormalComponentWrapper] | [juce::PopupMenu::HelperClasses::ItemComponent] | [juce::PopupMenu::HelperClasses::MenuWindow]
 
Instead of:
        // this | [juce::PopupMenu::HelperClasses::ItemComponent] | [juce::PopupMenu::HelperClasses::MenuWindow] 
as the code pretends to find.

Ask for my example code to me if you want to double check.


julianstorer:
Ok, thanks for the heads-up! I'll sort out a better fix for this..

PR #36, opened 2015-07-14
Fix shift-delete cut behavior in text editors

jkent:
Cut using shift-delete was broken in text fields due to delete taking precedence.


julianstorer:
Thanks - will sort this out!

PR #37, opened 2015-07-15
Linux HTTP enhancements

jkent:
I'd like some feedback on the way I implemented chunked encoding -- perhaps there is a better way I could have done this?


julianstorer:
Thanks! We'll have a look at this today and probably merge something in.

As you probably noticed, you sent us this on the same day that we added support for doing HTTP via libcurl! Since most people people will have curl installed that makes this old code largely redundant, but as a fallback we'll be keeping it around, and this certainly seems like a good addition!

PR #38, opened 2015-08-08
Imprvements to plugin host

FigBug:
Changes to FileBasedDocument
- send changed() when file name is updated

Changes to Plug-in Host:
- Add new document command
- Save open/closed state of each type of plug-in windows
- Save position for each type of plug-in window
- Open last used file on program launch
- Add current file name to application title bar
- Add command to bring all plug-in windows to front
- Added projects for new VS versions, re-saved jucer file


bazrush:
On 8 Aug 2015, at 18:43, Roland Rabien <notifications@github.com> wrote:
> Save open/closed state of each type of plug-in window
> Open last used file on program launch

Awesome ... there's an hour a month saved :)

PR #40, opened 2015-09-16
Fix svg parser failure on <path> element

kbinani:
`SVGState::parsePathString` will fail when two or more floating-point value(<1) exist in &lt;path&gt; element.
For example, &lt;path d="M.1.1L.9.9.1.9z"/&gt; should be parsed as:

```
Path path;
path.startNewSubPath(0.1, 0.1);
path.lineTo(0.9, 0.9);
path.lineTo(0.1, 0.9);
path.closeSubPath();
```

### DIFF
--- a/modules/juce_gui_basics/drawables/juce_SVGParser.cpp
+++ b/modules/juce_gui_basics/drawables/juce_SVGParser.cpp
@@ -1380,17 +1380,17 @@ Path Drawable::parseSVGPath (const String& svgPath)
 class SVGStateTests : public UnitTest
 {
 public:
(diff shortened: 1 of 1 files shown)


kbinani:
An example of valid svg:

https://cdn.rawgit.com/google/material-design-icons/master/image/svg/production/ic_edit_48px.svg

The rendered result by JUCE:
rendered_by_juce

julianstorer:
Thanks very much - we can't take pull requests directly, but will fix this shortly!

kbinani:
Thanks!

PR #42, opened 2015-10-16
Introjucer xcode custom plist fix (disable retina display)

magillo:
Function 
```sh
Introjucer/Source/Utility/jucer_MiscUtilities.cpp/addPlistDictionaryKey()
```
checks for existing xml child key before adding a new key and skips in case the key already exists.
This enable merging of Introjucer generated info.plist with custom xml properties for Xcode projects.
The  "already existing key" check snippet was missing from addPlistDictionaryKeyBool and addPlistDictionaryKeyInt, resulting in errors when adding bool or int values to custom xml in Introjucer.

e.g. if you try to disable retina display by adding 
```sh
<key>NSHighResolutionCapable</key> <false/>
```
The NSHighResolutionCapable (being a bool value) is added twice to info.plist xml, one with false value, one with default true value.
This confuses Xcode that uses the last (default) true value, actually ignoring the custom plist and always enabling retina display.


julianstorer:
Thanks - we added an equivalent (but more D.R.Y!) change for this last week.

PR #43, opened 2015-11-20
fix crash when window message lParam is 0 (Windows)

bome:
I cannot reproduce this directly, but a crash has happened with lParam=0.


julianstorer:
Thanks Florian - doesn't do any harm to add this kind of checking, I've pushed some equivalent code to do this now.

PR #45, opened 2015-11-26
mingw with win audio codecs

kluete:
IMHO all the namespace hoop-jumping in JUCE should be cleaned up
FYI on juce forums I'm user <petah>


modules/juce_audio_devices/juce_audio_devices.cpp
```diff
@@ -164,6 +164,12 @@ namespace juce
 
 //==============================================================================
 #elif JUCE_WINDOWS
+
+ #if JUCE_MINGW
+  } // namespace juce
```
julianstorer: This would allow all the smart pointer and message window code that follows to leak into the global namespace.. (?)

modules/juce_core/native/juce_win32_ComSmartPtr.h
```diff
@@ -30,11 +30,12 @@
...
 #if ! (defined (_MSC_VER) || defined (__uuidof))
-template<typename Type> struct UUIDGetter { static CLSID get() { jassertfalse; return CLSID(); } };
-#define __uuidof(x)  UUIDGetter<x>::get()
+ #warning "warning: disabling __uuidof()"
+ template<typename Type> struct UUIDGetter { static CLSID get() { jassertfalse; return CLSID(); } };
+ #define __uuidof(x)  UUIDGetter<x>::get()
 #endif
 
-inline GUID uuidFromString (const char* const s) noexcept
+inline GUID uuidFromString (const char* const s) noexcept		// how is this NOEXCEPT ???
```
julianstorer: ...huh? It's noexcept because nothing in that function throws an exception.

modules/juce_audio_formats/codecs/juce_WindowsMediaAudioFormat.cpp
```diff
@@ -22,6 +22,304 @@
...
+	WMT_RIGHT_COPY_TO_SDMI_DEVICE		= 0x10,
+	WMT_RIGHT_ONE_TIME			= 0x20,
+	WMT_RIGHT_SAVE_STREAM_PROTECTED		= 0x40,
+	WMT_RIGHT_COPY				= 0x80,
+	WMT_RIGHT_COLLABORATIVE_PLAY		= 0x100,
+	WMT_RIGHT_SDMI_TRIGGER			= 0x10000,
+	WMT_RIGHT_SDMI_NOMORECOPIES		= 0x20000
+} WMT_RIGHTS;
+    
+#include "C:/Program Files (x86)/Windows Kits/8.1/Include/um/nserror.h"
```
julianstorer: Really!?

julianstorer:
Thanks for the pull request - I've added a couple of quick comments about things that jumped out at me. Seems like it'd take a bit of cleaning up and testing for us to do this, but will try to find time if we can.

modules/juce_audio_formats/codecs/juce_WindowsMediaAudioFormat.cpp
kluete: well that's embarrassing :/ no sure why that was needed, possibly one of the MS files that mingw couldn't legally bundle but that might have changed by now. Let me know if/how I can help.

julianstorer: So does this it work if we just add the WMT_RIGHTS enum and nothing else?

kluete: will check asap, am in Los Angeles right noe

Sukumi: Many stuff has been redefined, especially the Media Format SDK stuff. But also inside the device API are types redefined which are available through MINGW MIDL or header files. I think several stuff due to historical reasons.
I agree that the marker/definition  forof feature, usage and actors(user) makes sense but the MINGW area has not been  resp. is not a focal area for past and future development. So, the effort to give mingw more coverage seems to be more driven by the engaged user base and naturaly less systematik as desired. 
So with the provisioned changes here and in the past it is sometimes not clear wether a mingw  marker/definition in code is a feature, a user or usage definition. Hard to achieve with one definition (JUCE_MING) anyway. Anyway, thanks for that nice piece of code and the opportunity to play around with it. cheers

hogliux:
Have you fixed the issues that Jules has mentioned? Otherwise we will close this PR.

kluete:
nope I haven't worked on audio apps for a year...

-- p

>Have you fixed the issues that Jules has mentioned? Otherwise we will close this PR.
>

PR #49, opened 2016-01-03
Inverse FFT method was calling forward FFT method

trevorknight:
The public FFT method `performRealOnlyInverseTransform (float*)`
should check for necessary scrach memory then call the private method
`performRealOnlyInverseTransform (Complex* scratch, float* d)`.
In one code branch, however, the forward transform was being called 
and this PR addresses that issue.


julianstorer:
Thanks for the heads-up, Trevor! Will sort that out..

trevorknight:
Thank you!

PR #50, opened 2016-01-04
failing aumu scan when creating a midi effect

pslack:
Here's why I did this

In our case we would like to have our AU plugin report as both aumu and aumi, the 2 are not mutually exclusive

When you create an AU midi effect plugin both aumi and aumu show up in the auval scan, with this fix both aumu and aumi can pass the validation.  Without this the aumu validation fails. 

--------------------------------------------------
RENDER TESTS:

ca_require: GetElement(inScope, inElement) != NULL InvalidElement /Users/wavedna/NewJucePlugin/JuceLibraryCode/modules/juce_audio_plugin_client/AU/CoreAudioUtilityClasses/AUBase.cpp:562
ERROR: -10877 IN CALL Get Output Format


julianstorer:
That code is only there for backwards-compatibility with legacy code, and we'll eventually deprecate it. Since we added the new bus arrangement functionality, you can just create whatever buses you need for your plugin, you don't need to use this code to do it.

pslack:
I don't need audio buses for my current project but I do need it to behave
like an aumu .. but the next project I will need multibus .. .. At the
moment I'm trying to stabilize my product build for release and to also
include some of the new JUCE features.

All very important changes for me thanks for all the great new feature
releases

julianstorer:
My point was that instead of hacking JUCE to re-enable those lines of code that add a bus, you can just do the same thing in your own class. That's how we intend people to use it in future anyway.

PR #51, opened 2016-01-13
Added push method to array class in juce_Javascript.cpp

audioplastic:
Added push method to array class


julianstorer:
Close, but no cigar! 
push actually takes a variable number of arguments, so I've implemented it differently - have a go and see if my version works for you.

PR #52, opened 2016-01-19
Remove extra ';'

monocasual:
I'm compiling a subset of JUCE modules for VST hosting purposes, with gcc version 5.3.1 on Linux (Debian stretch). This PR fixes warnings/errors about some extra semicolons here and there when JUCE modules are compiled with `-Wall -pedantic -Werror`.


julianstorer:
Thanks!

PR #53, opened 2016-01-26
Added comments and support for various textures/colors

drmadder:
- I added A LOT of comments for the newbies who are not familiar yet
  with OpenGL nor even completely JUCE-fluent to make it easier and more
  importantly much faster to comprehend at first sight the
  OpenGLAppExampleDemo
- changed "shader" variable name to "usedShaderProgram" for clarity sake
- changed frustum view to see a smaller object (teapot) without using
  scale local constant
- created a matrix-independent method (createRotationMatrix()) to
  create a rotation matrix instead of making it a member of a matrix
  class, which is not necessary and confusing. It actually should be a
  static method in the Matrix3D class but I did not want want to modify
  the module files myself
- added #define constants and code to compile a version with the
  existing color (generated in the shader), one with a fixed color
  (defined in the C++ code) or one with a texture file (stored as
  BinaryData)
- renamed "texureCoordIn" to "textureCoordIn"
- renamed "newShader" to "newShaderProgram"
- added a texture (jpg) file resource to be added in the BinaryData
  files

I did not modify any module nor the WavefrontObjParser.h file, which
would still deserve more comments, but is not necessary to understand
in a first approach.


hogliux:
Hey drmadder,
wow this looks really nice and very well documented. Also thank you for some of the minor fixes - I've added them to develop. 

The documentation is super valuable for anybody who wants to start on OpenGL and JUCE. I think it may be too "educational" for in-line comments in the juce demo code. We don't really do this sort of documentation with in-line comments.

I think a tutorial is a better place for this. We are planning to release a JUCE & OpenGL tutorial soon and I think this would be a great starting point. Is it ok for us to use your comments and project for future tutorial?

Fabian

PR #54, opened 2016-02-07
Add basic log string and supporting debug mode macro

g40:
Hello Jules,

This is a simplified string stream class and supporting macro specifically for debugging/tracing. Long story short:

DBM("Hello World " << 42 << ':' << true);

ends up piping 

path\to\source\dlgpassword.cpp(153) : Hello World 42 true to juce::Logger::outputDebugString()

Implementation is both thread and type safe with macro (optionally) cleanly disabled in non-debug builds.

This implementation builds off of the existing juce::String for various conversions. The set of various overloaded operator<<() functions is not yet complete, as I'm not sure if this might be of interest. 

My personal motivation for adding is to simplify tracing what I think is a focus issue in modal dialogs (example to follow)

BTW, no idea about copyright for these proposals - does all get assigned to ROLI?

BR.

Jerry

### DIFF
--- a/modules/juce_core/juce_core.h
+++ b/modules/juce_core/juce_core.h
@@ -194,6 +194,7 @@ extern JUCE_API void JUCE_CALLTYPE logAssertion (const char* file, int line) noe
 #include "text/juce_String.h"
 #include "text/juce_StringRef.h"
 #include "logging/juce_Logger.h"
+#include "logging/juce_LogString.h"
 #include "memory/juce_LeakedObjectDetector.h"
 #include "memory/juce_ContainerDeletePolicy.h"
 #include "memory/juce_HeapBlock.h"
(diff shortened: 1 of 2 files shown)


julianstorer:
..or in your own app just define
#define DBM(x)  DBG (__FILE__ << "(" << __LINE__ << ") : " << x)
?

Yes, if we ever directly accepted a pull request from an external source, we would need to get a written agreement to take over the copyright. But so far nobody's ever submitted anything good enough in style, idea and execution that we've even considered that!

If you're keen to contribute, that's awesome, but maybe better to discuss the problem on the forum first, because if I was to implement this feature, it wouldn't resemble what you've got here.. (Also read our coding standards guide!)

g40:
OK. Each to their own :)

To clarify I've got a bug/issue here best illustrated with tiny project. Do you want it via an issue/PR here or something uploaded to Juce forum?

Thx.

julianstorer:
The forum would be best, cheers!

PR #55, opened 2016-02-16
fix typo in File class documentation

bill-auger:
the documentation for the File::File() constructor reads:

> "... so getFullPath() will return an empty string"

but there is no such method - prbably referring to getFullPathName() - yes?


julianstorer:
Thanks Bill, will tidy that up!

PR #56, opened 2016-02-22
Fix HAVE_STDLIB_H redefinition

monocasual:
This PR tries to fix a macro redefinition with GCC 5.3.1 on Debian testing (Linux), when compiled with `-Werror` flag.


julianstorer:
Seems pretty obscure, but thanks, I'll sort that out!

PR #57, opened 2016-03-02
MSVC compiler warning C4512 fixed

VioletGiraffe:
MSVC warning C4512 "assignment operator could not be generated" fixed

### DIFF
--- a/modules/juce_audio_plugin_client/utility/juce_PluginBusUtilities.h
+++ b/modules/juce_audio_plugin_client/utility/juce_PluginBusUtilities.h
@@ -27,6 +27,8 @@ struct PluginBusUtilities
 {
     //==============================================================================
     typedef Array<AudioProcessor::AudioProcessorBus> AudioBusArray;
+    
+    PluginBusUtilities& operator=(const PluginBusUtilities&) = delete; // To avoid "assignment operator could not be generated" compiler warning
 
     //==============================================================================
     PluginBusUtilities (AudioProcessor& plugin, bool markDiscreteLayoutsAsSupported)


julianstorer:
Huh? That class already has a JUCE_DECLARE_NON_COPYABLE at the bottom, which adds exactly the same operator= definition..

VioletGiraffe:
Hm. Didn't notice that macro. Don't know what's wrong, but I do get the warning out of the box and don't get it with my proposed edit.

julianstorer:
Sorry, but we don't get that warning here, and your change should cause a compiler error because it's multiply-defining the same operator.. No idea what you could be doing that enables your change to compile!

VioletGiraffe:
It seems that `JUCE_DECLARE_NON_COPYABLE` actually has no effect, i. e. isn't defined to what it should be (despite Visual Studio displaying it as defined properly). I'll look into it.

VioletGiraffe:
Actually, class `PluginBusUtilities` does NOT have the `JUCE_DECLARE_NON_COPYABLE` macro, plain and simple. hence this class does produce the warning.

julianstorer:
Yes it does. 
You must be looking at some other version of the file.

julianstorer:
Just to end this discussion:
https://github.com/julianstorer/JUCE/blob/master/modules/juce_audio_plugin_client/utility/juce_PluginBusUtilities.h#L385

VioletGiraffe:
I see. It's not present in the version currently on your main site, but it is here in the repository. I had to download JUCE from the site because the repo lacks Projucer.

PR #58, opened 2016-03-09
don't call method on null object

kluete:
don't post message to null msg manager, whose instantiation isn't immediate


julianstorer:
Thanks, I'll add a nullptr check for that!

PR #59, opened 2016-03-11
Fix missing field initializer

dawikur:
GCC is complaining for missing initializer. I am assuming the intended action  here is default construction. Passing `{}` makes compiler happy.

### DIFF
--- a/modules/juce_gui_basics/drawables/juce_SVGParser.cpp
+++ b/modules/juce_gui_basics/drawables/juce_SVGParser.cpp
@@ -905,7 +905,7 @@ private:
                 }
             };
 
-            GetFillTypeOp op = { this, &path, opacity };
+            GetFillTypeOp op = { this, &path, opacity, {} };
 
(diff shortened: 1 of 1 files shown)


julianstorer:
Thanks, I'll take a look. AFAIK the syntax that's there is correct, so GCC might be moaning unnecessarily, but will find a way to keep it happy..

dawikur:
Checked in n4567

> § 8.5.1
> 7
> If there are fewer initializer-clauses in the list than there are members in the aggregate, then each member
> not explicitly initialized shall be initialized from its default member initializer (9.2) or, if there is no default
> member initializer, from an empty initializer list (8.5.4).

You are right, this is gcc bug. Should have checked in standard first (lesson learned). Sorry for bothering.

julianstorer:
No problem - thanks for flagging it. Even if it's a GCC bug then it's still
a good thing for us to add a workaround for this kind of thing.

PR #60, opened 2016-03-12
Add c++14 as a valid entry for C++ standard to use

dawikur:


julianstorer:
Thanks! Good idea, we'll add something to do that!

PR #62, opened 2016-03-21
Split out JuceAppStartupDelegate to header file

adamski:
This pull request is to make the `window` property available from `JuceAppStartupDelegate`, as it is on the standard iOS `AppDelegate`. This is necessary for some iOS libraries to function that expect the window property on the current `UIApplication` instance.

##### Usage
I have a custom `MainWindow` Objective-C++ class that creates the initial `UIWindow` instance:
```Objective-C++
#import "<Path-to-JUCE>/modules/juce_gui_basics/native/juce_ios_Windowing.h"
//...
MainWindow::MainWindow (String name)  : DocumentWindow (name,
                                                        Colours::lightgrey,
                                                        DocumentWindow::allButtons)
{
    //...
    JuceAppStartupDelegate *appDelegate = (JuceAppStartupDelegate *)[[UIApplication sharedApplication] delegate];

    appDelegate.window = [[UIWindow alloc] initWithFrame:[[UIScreen mainScreen] bounds]];
    appDelegate.window.backgroundColor = [UIColor whiteColor];
    // ...
}
```

It would be convenient and cleaner to have the new header file in `JuceHeader.h` but given the rarity of potential usage maybe its ok to include it from the JUCE modules folder.

### DIFF
--- a/modules/juce_gui_basics/native/juce_ios_Windowing.h
+++ b/modules/juce_gui_basics/native/juce_ios_Windowing.h
@@ -1,0 +1,21 @@
+#ifndef juce_ios_Windowing_h
+#define juce_ios_Windowing_h
+
+@interface JuceAppStartupDelegate : NSObject <UIApplicationDelegate>
+{
+}
+
+@property (strong, nonatomic) UIWindow *window;
+
+- (void) applicationDidFinishLaunching: (UIApplication*) application;
+- (void) applicationWillTerminate: (UIApplication*) application;
+- (void) applicationDidEnterBackground: (UIApplication*) application;
+- (void) applicationWillEnterForeground: (UIApplication*) application;
+- (void) applicationDidBecomeActive: (UIApplication*) application;
+- (void) applicationWillResignActive: (UIApplication*) application;
+
+@end
+
+
+
+#endif /* juce_ios_Windowing_h */
--- a/modules/juce_gui_basics/native/juce_ios_Windowing.mm
+++ b/modules/juce_gui_basics/native/juce_ios_Windowing.mm
@@ -35,20 +35,10 @@ Array<AppInactivityCallback*> appBecomingInactiveCallbacks;
 
 } // (juce namespace)
 
-@interface JuceAppStartupDelegate : NSObject <UIApplicationDelegate>
-{
+#import "juce_ios_Windowing.h"
+
+@implementation JuceAppStartupDelegate {
 }
-
-- (void) applicationDidFinishLaunching: (UIApplication*) application;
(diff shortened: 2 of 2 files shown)


hogliux:
I don't entirely understand your sample code in your comment. Why do you need to cast this to JuceAppStartupDelegate? Surely you could just cast it to NSObject<UIApplicationDelegate> and do the same?

adamski:
But that still leaves the problem of needing to access the `window` property of the object. Are you suggesting I could dynamically add the window property there (to an NSObject) instead of declaring it in JuceAppStartupDelegate?

adamski:
The requested change is necessary for me to use React Native with JUCE, and it is also expected of some other iOS libraries I've seen. If there is an alternative way to achieve this I would be happy to know.

hogliux:
The window property is already in the base class: see this [link from the Apple docs.](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIApplicationDelegate_Protocol/#//apple_ref/occ/intfp/UIApplicationDelegate/window)

adamski:
JuceAppStartupDelegate does not have that property, see https://forum.juce.com/t/adding-uiwindow-to-juceappstartupdelegate/16519  - a library I am now using needs to access this property also. That is why I added it to the class. Unless I am missing something here?

hogliux:
Yes but the base class does have this property. If you do not override this property the base class should store it for you. What happens if you compile this:

```
MainWindow::MainWindow (String name)  : DocumentWindow (name,
                                                        Colours::lightgrey,
                                                        DocumentWindow::allButtons)
{
    //...
    NSObject<UIApplicationDelegate> *appDelegate = [[UIApplication sharedApplication] delegate];

    appDelegate.window = [[UIWindow alloc] initWithFrame:[[UIScreen mainScreen] bounds]];
    appDelegate.window.backgroundColor = [UIColor whiteColor];
    // ...
}
```

adamski:
Thanks I'll try that out and let you know/

adamski:
@hogliux, thank you, that compiles fine. And thanks for the Objective-C schooling I didn't realise this was possible. I'l resurrect that thread and post the answer too.

adamski:
@hogliux I'm not sure why, but I didn't test this properly. I was still using my branch of JUCE. So, I still have this issue. The last code you posted fails with 
```Objective-C
[JuceAppStartupDelegate setWindow:]: unrecognized selector sent to instance
```

I believe this is because the JuceAppStartupDelegate class needs to declare

```Objective-C
@property (strong, nonatomic) UIWindow *window;
```

adamski:
Realised that using the code you suggested means all thats needed is one line, see #86

PR #63, opened 2016-03-24
Fix for repaint of components that are cached and scaled

wkundrus:
The test for valid areas needs to use the scaled image bounds.

### DIFF
--- a/modules/juce_gui_basics/components/juce_Component.cpp
+++ b/modules/juce_gui_basics/components/juce_Component.cpp
@@ -792,7 +792,7 @@ public:
             validArea.clear();
         }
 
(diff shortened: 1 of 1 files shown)


julianstorer:
Hmm.. No, I don't think that's correct. 
The validArea isn't scaled, and lower down the function it gets set to the compBounds, not imageBounds. Maybe email me some example code that demonstrates the problem and I'll see if I can find a better fix?

wkundrus:
Ah, I see, it had been fixed in invalidate(); It used to look like this:

 bool invalidate (const Rectangle<int>& area) override    { validArea.subtract (area * scale); return true; }

PR #64, opened 2016-03-26
seebk QtCreator exporter

rsenn:


julianstorer:
Thanks! Adding QtCreator support is something we'd like to do, and this is probably a great start, but a couple of problems:
- If anyone's contributing more than a couple of lines we'd need to sort out a licensing agreement so that we own copyright (we can't just accept a pull request)
- This pull request contains 20,000 changes and has merge conflicts, so it's hard to even see what's going on! 

Also, next week we're going to be heavily restructuring the introjucer folder, so best plan is probably to wait until that, and then do a new clean pull request based on that..

PR #66, opened 2016-04-01
Correction in dilution factor.

Skilpad:
Each dilution (1C) is a dilution by a factor 100, not 10.


julianstorer:
Drat, my homeopathy professor will be so disappointed! 
Thanks for letting us know - getting a basic thing like that wrong could have caused the whole thing to stop working!
(I'd love to accept your pull request but we actually can't accept them on this JUCE repo..)

PR #68, opened 2016-04-07
Markdown formatting

adamski:
Rename `.txt` files to '.md' and tidy up formatting for the JUCE Module Format doc.


julianstorer:
Thanks Adam, but we don't really want these files to be markdown.. They need to use a format that you can open absolutely anywhere, and markdown just isn't supported widely enough. Even on mordern OSes like OSX it won't show a quick preview of a .md file, annoyingly!

adamski:
Fair enough, I understand - I just like nicely formatted docs to read :)

PR #69, opened 2016-04-10
Fixed code and comment spelling errors.

mtytel:
These errors were found using 
`$ codespell --quiet-level=3`

I'm not this meticulous.

I may have corrected some library files as well.
If you just want JUCE specific corrections I can make a different pull request.


julianstorer:
Thanks Matt, didn't realise there were this many typos! We'll need to avoid changing anything in the 3rd party libraries as that'll make it harder to merge in changes when they get updated, but we'll sort out the mistakes in the juce files..

PR #72, opened 2016-04-21
fix build if not audio plugin

bome:
minimal change to fix build


hogliux:
This commit is not necessary anymore since this logic has been removed when we switched over to static shared library code.

PR #73, opened 2016-04-22
support lambdas for var::NativeFunction by using std::function

kunitoki:
Only if we have a c++11 compiler. Should work ok in older compilers but obviously they will not support using lambdas.


hogliux:
Thanks. I've added this to JUCE.

kunitoki:
great thanks!

PR #78, opened 2016-05-06
Fix colour for toggled button in LookAndFeelV3

ServiusHack:
The monster commit 70949aa changed `juce_LookAndFeel_V3.cpp` to set `textButtonColour` for `TextButton::buttonOnColourId`, which is also used for `TextButton::buttonColourId`. This makes it impossible to distinguish the toggle state of buttons. Simply removing the line restores the previous behavior.


hogliux:
Simply removing this line is not a good solution. On the latest develop tip I've changed the colour to be different to buttonColoutId. Thanks for pointing this out.

PR #84, opened 2016-06-16
Fixed the code that determines the modifier masks on Linux.

thomasloockx:
The code that determines the modifier masks on Linux is broken. My patch fixes that. The old code seemed to assume that the modifiermap always has 2 entries per modifier key (i.e. max_keypermod == 2). That's an invalid assertion and max_keypermod should be used instead.

This is also discussed in the JUCE forum:
https://forum.juce.com/t/alt-key-not-working-on-linux/15830?u=tloockx


hogliux:
Hi thomasloockx,

I've pushed your fix to develop along with some other X11 fixes.

Fabian

PR #86, opened 2016-06-25
Add window property to JuceAppStartupDelegate

adamski:
In order to be able to use certain iOS libraries that expect a `window` property in the `UIApplication` instance, we need this to be declared in `JuceAppStartupDelegate`. 

We can then set it up as follows
```Objective-C
NSObject<UIApplicationDelegate> *appDelegate = [[UIApplication sharedApplication] delegate];
appDelegate.window = [[UIWindow alloc] initWithFrame:[[UIScreen mainScreen] bounds]];
appDelegate.window.backgroundColor = [UIColor blackColor];
```

### DIFF
--- a/modules/juce_gui_basics/native/juce_ios_Windowing.mm
+++ b/modules/juce_gui_basics/native/juce_ios_Windowing.mm
@@ -39,6 +39,8 @@ Array<AppInactivityCallback*> appBecomingInactiveCallbacks;
 {
 }
 
+@property (strong, nonatomic) UIWindow *window;
+
 - (void) applicationDidFinishLaunching: (UIApplication*) application;
 - (void) applicationWillTerminate: (UIApplication*) application;
 - (void) applicationDidEnterBackground: (UIApplication*) application;


hogliux:
OK I've added this to the develop branch. Thank you!

designerfuzzi:
was reading the Juce Forum post and thought i share one point that immediately jumped in my eyes.
Since a couple of years (since iOS 12)  `NSWindow *window;` is a property of `UISceneDelegate` and not a property of UIApplicationDelegate. 
Meaning you add the window property because you assume the old API is still valid. While this helps you maintaining very old code it doesnt allow you to properly manage Scenes. The idea is that one App can have multiple Windows which are managed by scenes adapting to multiple use cases. 

The workaround to add some **window** property to `UIApplicationDelegate` works because a UIWindow (even the mainWindow) just needs some pointer to dangle around and it doesnt matter where it does so. The only thing you want to avoid at all cost is destroying the main Window before your App quits.

adamski:
Thanks. Can you suggest a more modern alternative?

designerfuzzi:
the troubles come up because the iOS Xcode template doesnt include versioning Macros to address development to older systems.  Developers are responsible them self which versions they support. Also some classic UIApplicationDelegate methods are just not called starting version iOS13.. 

the property change was obligatory in iOS12 and is mandatory since iOS13.
if the versioning macros of Xcode/objective-c are used..<br>the UIApplicationDelegate looks something more like 

for the sake of clearance whats actually going on with SceneDelegate, some example stuff included here.
```objc
@interface YourAppDelegate : UIResponder <UIApplicationDelegate>
@property (strong, nonatomic) UIWindow *window
API_DEPRECATED_WITH_REPLACEMENT("SceneDelegate:window", ios(2.0, 13.0)) NS_DEPRECATED_IOS(2.0, 13.0);
@end
```
implementation side..
```objc
/// still called in iOS13.0+
-(BOOL)application:(UIApplication *)application willFinishLaunchingWithOptions:(nullable NSDictionary *)launchOptions
{
    //initialize your own stuff here..
    return YES;
}

/// still called in iOS13.0+ 
-(BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{
    
    if (@available(iOS 13.0, *)) {
        
        //on iOS13, mainViewController is set via SceneDelegate
        //also external displays with their own Scene, Window and ViewController
        
    } else {
        
        // Fallback for code earlier <iOS13.0
        _mainViewController = [[YourMainViewController alloc] init];
        self.window.rootViewController = _mainViewController;
        
        // alloc+init external displays here when used prior iOS13
        //_externalDisplay = [[SomeExternalDisplay alloc] init]; 
    }
    
    return YES;
}

/// NOT called in iOS13.0+
-(void)applicationDidBecomeActive:(UIApplication *)application {
    [self thinkTwiceCodeOnce_DidBecomeActive]; //fallback prior <iOS13.0
}
/// NOT called in iOS13.0+
-(void)applicationWillEnterForeground:(UIApplication *)application {
    [self thinkTwiceCodeOnce_WillEnterForeground]; //fallback prior <iOS13.0
}
/// NOT called in iOS13.0+
-(void)applicationDidEnterBackground:(UIApplication *)application {
    [self thinkTwiceCodeOnce_DidEnterBackground]; //fallback prior <iOS13.0
}

// former -applicationDidBecomeActive: code here 
// to be called from SceneDelegate starting iOS13.0
-(void)thinkTwiceCodeOnce_DidBecomeActive {
}
// former -applicationWillEnterForeground: code here 
// to be called from SceneDelegate starting iOS13.0
-(void)thinkTwiceCodeOnce_WillEnterForeground {
}
// former -applicationDidEnterBackground: code here 
// to be called from SceneDelegate starting iOS13.0
-(void)thinkTwiceCodeOnce_DidEnterBackground {
}

/// still called in iOS13+
- (void)applicationWillTerminate:(UIApplication *)application {
    
}

//mandatory since iOS13 but throws error prior, so mark it API_AVAILABLE 
- (UISceneConfiguration *)application:(UIApplication *)application configurationForConnectingSceneSession:(UISceneSession *)connectingSceneSession options:(UISceneConnectionOptions *)options
API_AVAILABLE(ios(13.0)) 
{
    // Called when a new scene session is being created.
    // Use this method to select a configuration to create the new scene with.
    // the new scene with. @"Default Configuration";
    UISceneConfiguration *configuration = [[UISceneConfiguration alloc] initWithName:@"Default Configuration" sessionRole:connectingSceneSession.role];
    return configuration;
}

//mandatory since iOS13 but throws error prior, so mark it API_AVAILABLE 
- (void)application:(UIApplication *)application didDiscardSceneSessions:(NSSet<UISceneSession *> *)sceneSessions
API_AVAILABLE(ios(13.0))
{
    // Called when the user discards a scene session.
    // If any sessions were discarded while the application was not running, this will be called shortly after application:didFinishLaunchingWithOptions.
    // Use this method to release any resources that were specific to the discarded scenes, as they will not return.
    //[UIApplication openSessions]
    
}
```
and the corresponding SceneDelegate
```objc
API_AVAILABLE(ios(13.0))
@interface SceneDelegate : UIResponder <UIWindowSceneDelegate>
@property (strong, nonatomic) UIWindow * window;
@end
```
implementation side
```
@implementation SceneDelegate
- (void)scene:(UIScene *)scene willConnectToSession:(UISceneSession *)session options:(UISceneConnectionOptions *)connectionOptions {
    // Use this method to optionally configure and attach the UIWindow `window` to the provided UIWindowScene `scene`.
    // If using a storyboard, the `window` property will automatically be initialized and attached to the scene.
    // This delegate does not imply the connecting scene or session are new (see `application:configurationForConnectingSceneSession` instead).
    
    UIWindowScene *windowScene = (UIWindowScene*)scene;
    _window = [[UIWindow alloc] initWithWindowScene:windowScene];
    
    if (_window.rootViewController==nil) {
        NSLog(@"weirdo! rootViewController still missing, create one");
        YourAppDelegate* appdelegate = (YourAppDelegate*)[[UIApplication sharedApplication] delegate];
        YourMainViewController *mainViewController = [[YourMainViewController alloc] init];
        _window.rootViewController = mainViewController;
    }
    
    [_window makeKeyAndVisible];
}

- (void)sceneDidDisconnect:(UIScene *)scene { //API_AVAILABLE(ios(13.0))
    // Called as the scene is being released by the system.
    // This occurs shortly after the scene enters the background, or when its session is discarded.
    // Release any resources associated with this scene that can be re-created the next time the scene connects.
    // The scene may re-connect later, as its session was not neccessarily discarded (see `application:didDiscardSceneSessions` instead).
    
}

- (void)sceneDidBecomeActive:(UIScene *)scene {
    // Called when the scene has moved from an inactive state to an active state.
    // Use this method to restart any tasks that were paused (or not yet started) when the scene was inactive.
    YourAppDelegate* appdelegate = (YourAppDelegate*)UIApplication.sharedApplication.delegate;
    [appdelegate thinkTwiceCodeOnce_DidBecomeActive];
}

- (void)sceneWillResignActive:(UIScene *)scene {
    // Called when the scene will move from an active state to an inactive state.
    // This may occur due to temporary interruptions (ex. an incoming phone call).
    YourAppDelegate* appdelegate = (YourAppDelegate*)UIApplication.sharedApplication.delegate;
    [appdelegate thinkTwiceCodeOnce_DidBecomeActive];
}

- (void)sceneWillEnterForeground:(UIScene *)scene {
    // Called as the scene transitions from the background to the foreground.
    // Use this method to undo the changes made on entering the background.
    YourAppDelegate* appdelegate = (YourAppDelegate*)[[UIApplication sharedApplication] delegate];
    [appdelegate thinkTwiceCodeOnce_WillEnterForeground];
}

- (void)sceneDidEnterBackground:(UIScene *)scene {
    // Called as the scene transitions from the foreground to the background.
    // Use this method to save data, release shared resources, and store enough scene-specific state information
    // to restore the scene back to its current state.
    YourAppDelegate* appdelegate = (YourAppDelegate*)[[UIApplication sharedApplication] delegate];
    [appdelegate thinkTwiceCodeOnce_DidEnterBackground];
}
@end
```
So the upper example marks properties to be not implemented or offers some entry points to keep the code almost at place, not coding twice.. Meaning its possible to support iOS12 & iOS13 with the same code base despite this is kinda obsolete in near future.

PR #87, opened 2016-07-16
Fix Audio Unit PropertyChangeListener

x42:
mPropertyID was hardcoded overridden.


hogliux:
Wow! What an embarrassing typo. Thanks for pointing this out. It's now fixed on the development branch.

PR #104, opened 2016-08-25
Android bridge

adamski:
Separate `JuceAppActivity` JNI interactions and JUCE specific functionality into separate `JuceBridge` class. Separate `ViewHolder` inner class into separate `JuceViewHolder`.

Modify Android exporter, keeps java files in `com.juce` package, allowing user to copy and modify the activity without disturbance from Projucer. 

See [forum post](https://forum.juce.com/t/modifications-to-juce-android-implementation/18860).


adamski:
Just in case you were going to look at this, I'm currently in the process of redoing some of it to implement`Component::addToDesktop` for Android, and remove the need to store a `Map` of ViewHolders.

adamski:
I actually scrapped the `Component::addToDesktop` implementation for now - led me into some obscure JNI errors, I think due to the View reference being cast to `void*`. 

Some improvements to the Java side, ComponentPeerView's are stored in a HashMap rather than ViewHolders.
The current implementation is compatible with React Native v0.3.1

hogliux:
Due to the size of this PR, I'm taking the discussion offline. I've sent you a PM on the forum.

PR #108, opened 2016-09-13
Improve graph editor panel in audio plugin host

harry-g:
## a very slight improvement for the audio plugin host demo:
- Currently, if you double-click a plugin with no parameters, an empty dialog appears.
  - E.g. try the internal plugins audio in/out and midi in/out.
- With the change, a plugin without UI and without parameters will not open up any dialog.


julianstorer:
I think it's probably less confusing for a user to get an empty dialog than to get no response at all when they're expecting a window to appear..?

harry-g:
If you think so... for me it was confusing to get an empty dialog...

PR #124, opened 2016-11-07
Some fixes from the Debian packaging of juce

umlaeute:
here's a number of patches from the Debian  ['juce'](https://tracker.debian.org/pkg/juce) package.

# fix building with system-provided libraries
- the `z_uInt` and `png16` patches are needed to build juce without the bundled libraries.
  (please note that the `png16` patch has not really been tested against compatibility with older libpngs)

# CPP/C/CXX/LDFLAGS
it's great that juce now uses `JUCE_CFLAGS` (and similar) instead of `CFLAGS`.
it's even better that you include `CFLAGS`, so the build-process can inject additional flags.
however, it would be best if those flags were *appended* rather than *prepended*, as this would allow to override some of these FLAGS (provided the builder knows what they are doing).

e.g. if the builder policy is to build all packages with `-O0`, the usual way would be to run

    make CFLAGS="-O0"

however, since currently the CFLAGS are prepended,  this could expand to some actual compiler flags like `[...] -O0 [...] -fPIC -O3`, effectively disabling the builder's `-O0` flag (since later flags of the
 same kind will override earlier one, in this case `-O3` is used)

# silent builds
using the "@" prefix to silence builds is often great.
sometimes it can be annoying, e.g. when running automated builds on remote server where we don't have direct access to (think: continuous integration).
    
this patch uses the logic from autotools silent builds, which can be disabled when run as 

    make V=1

(`V=1` is used by autotools; CMake uses `VERBOSE=1`)

# reproducible builds
in Debian we would like to make sure that building the source code multiple times, gives the same results **bit for bit**.
This obviously won't work if the resulting binaries include things like the current date (at time of building).
Therefore we would like to include a patch that allows overriding of the current build time from the cmdline.

See also https://reproducible-builds.org/

### DIFF
--- a/extras/Projucer/Source/Application/jucer_Application.cpp
+++ b/extras/Projucer/Source/Application/jucer_Application.cpp
@@ -130,6 +130,9 @@ void ProjucerApplication::initialiseBasics()
     icons = new Icons();
 }
 
+#ifndef BUILD_DATE
+# define BUILD_DATE __DATE__
+#endif
 bool ProjucerApplication::initialiseLogger (const char* filePrefix)
 {
     if (logger == nullptr)
@@ -142,7 +145,7 @@ bool ProjucerApplication::initialiseLogger (const char* filePrefix)
 
         logger = FileLogger::createDateStampedLogger (folder, filePrefix, ".txt",
                                                       getApplicationName() + " " + getApplicationVersion()
-                                                        + "  ---  Build date: " __DATE__);
+                                                        + "  ---  Build date: " BUILD_DATE);
         Logger::setCurrentLogger (logger);
     }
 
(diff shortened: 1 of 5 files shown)


hogliux:
I've reviewed your changes, cleaned them up and pushed most of them to develop. Thank you!

I didn't yet merge in the build-time changes as I think that `Time::getCompilationDate` should do what it says on the box and there should be no way to override this. However, it might be useful to have a flag that makes the Projucer not print the compile time on app startup.

umlaeute:
thanks for the merge.
regarding the don't-print-the-date-flag: this is actually not helpful for the reproducible build effort, because the build-date will still be in the binary (but with less actual uses).

anyhow, it seems that `gcc` (at least as packaged in Debian) supports setting of the "current time", which makes my patch against juce less important (as the problem is fixed elsewhere)

PR #125, opened 2016-11-09
added text drag outside of app for mac

genia:


genia:
Integrated Cocoa code from:
https://developer.apple.com/library/content/samplecode/CocoaDragAndDrop/Introduction/Intro.html

julianstorer:
Thanks, I've commited a DRYer implementation that should achieve the same thing - let me know if you have any problems with it.

PR #126, opened 2016-11-13
Projucer Xcode exporter "keep custom targets" and AUv3/Standalone minimum SDK

wolframfranke:
1. Added a new boolean property labelled "Keep custom Xcode schemes" to the Projucer which keeps existing schemes with custom debugger settings or launch executables for plugin development.

2. AUv3 and stand-alone plug-in targets are automatically set upon export to use the Mac OS 10.11 SDK if an older SDK has been set for other targets.


hogliux:
Thank you! I've added this to the develop branch.

PR #127, opened 2016-11-15
Develop

jamiebullock:
This PR removes the flag INTERNET_FLAG_NO_COOKIES from the flags passed to `HttpOpenRequest()`.

According to the WinInet API documentation 

> **INTERNET_FLAG_NO_COOKIES** 
> Does not automatically add cookie headers to requests, and does not automatically add returned cookies to the cookie database

However, this behaviour is inconsistent with the implementation of WebInputStream on the Mac platform, which does automatically add cookie headers. This change makes the behaviour of `openHTTPConnection()` more consistent between platforms.


julianstorer:
Sorry, but I can't see which changes in here are yours, as you seem to have also dragged hundreds of other changes in along with them.. Maybe send us another PR against the latest version of develop so we can tell what you're asking for?

PR #128, opened 2016-11-25
Improve capability to check whether a socket is connected locally

hm1992:
Current check for if a socket is connected locally is (hostname == "127.0.0.1"). This is not complete as we could have connected to any local IP address, local hostname or "localhost" as well.

This pull request checks the ip address we connected to (regardless of how we got there) against all local IP addresses.

Also it adds an internal function which retrieves the IPv4 or IPv6 (future proof) string from the sockaddr struct.

### DIFF
--- a/modules/juce_core/network/juce_Socket.cpp
+++ b/modules/juce_core/network/juce_Socket.cpp
@@ -327,11 +327,34 @@ namespace SocketHelpers
         return nullptr;
     }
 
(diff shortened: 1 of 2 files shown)


hm1992:
This pull request is fixed by https://github.com/julianstorer/JUCE/commit/6340e541f08e0890e08915363a207a8a8091c1d2

PR #133, opened 2016-12-05
Fix for an issue with PropertiesFile when applicationName contains a dot

francoisbecker:
.withFileExtension(filenameSuffix) did erase the part after the last dot contained in applicationName, as if this part after the last dot was already a file extension.

Instead, the extension is just appended.


francoisbecker:
Closed as it has been fixed in the develop branch

PR #138, opened 2016-12-10
Copy Paste

jcredland:
Jules/All - this adds a feature where paths can be copied from specific modules, to other selected modules.  See picture for details. 

This makes working with non-juce standard modules much easier, where not all modules need the paths set the same

image


julianstorer:
Thanks Jim! Have added this now.

jcredland:
Ta!  Much appreciated.

PR #142, opened 2016-12-14
Projucer: added a setting for the maximum number of recent items…

francoisbecker:
…to be displayed in the “Open Recent” menu; added a “Misc” settings panel for this new setting.

Also, fixed a small doc typo in extras/Projucer/Source/Application/jucer_GlobalPreferences.h

### DIFF
--- a/extras/Projucer/Source/Application/jucer_GlobalPreferences.cpp
+++ b/extras/Projucer/Source/Application/jucer_GlobalPreferences.cpp
@@ -363,11 +363,48 @@ void AppearanceSettingsTab::resized()
 }
 
 //==============================================================================
+MiscSettingsTab::MiscSettingsTab()
+{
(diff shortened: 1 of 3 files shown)


julianstorer:
Thanks Francois, we appreciate the effort, but adding a new page for just this one setting seems a bit like overkill considering how minor the feature is! Can you think of a more minimal way we could do this? (Personally I'd be happy to just bump up the number, but it seems that other people don't want that..)

francoisbecker:
Yes, this is overkill (read my words on https://forum.juce.com/t/patch-projucer-max-number-of-recent-items/20256 ), except if the there is no other solution and nobody agrees on the right number of items: this is not "overkill" anymore but the "best solution atm".

Anyway thanks for having answered about the possibility (not) to have it in JUCE at the moment, this way I know I can take advantage of adding this on my fork, instead of having to manage a merge between your rewrite of the patch and my initial write of the patch.

PR #149, opened 2017-01-19
fix toggle button text not visible with default colours in LookAndFeel_V3

harry-g:
The toggle button text is not visible with default colours in LookAndFeel_V3.
The background for button on is black, default text colour also.
See the JUCE Demo -> Look&Feel -> Select LookAndFeel_V3 and buttons in toggle mode (setClickingTogglesState(true)).

Therefore, instead of black, we should have
`    setColour (TextButton::buttonOnColourId, Colours::grey);`


julianstorer:
Cheers, I'll sort that out

PR #150, opened 2017-01-19
check for return value of writeMessage and writeBundle in OSC send

harry-g:
writeMessage and writeBundle can fail for various reasons (~3 as seen in the source code).
In OSC send, their return value is not yet checked.
I think it would be safer to check it.


julianstorer:
Thanks Harry, I'll get that fixed today

PR #152, opened 2017-01-20
Improvements to ProJucer

gbevin:
This contains a series of improvements to ProJucer:

* Added support for dynamic paint code.
The code generated by the paint objects, now generates variables for x, y, width, height and the different colors. This are put together with the paint code in a dedicated block for an isolation scope. After the variables declaration and definition, a comment is insert that allows developers to modify the values of those variables. Since there's an arbitrary number of comment blocks possible, they're handled differently from the other ProJucer comments, in that the same name is reused. The individual comment is identified by the order of occurrence, which matches the order of the paint objects. When the code is parsed, the paint objects receive a copy of the customization code, allowing it to be included when the code is generated. The standard way of copying over code inside comment blocks from an previous version, now has a bypass condition that make it ignore these new comments.


julianstorer:
This sounds great, but I don't really trust this merge as it seems to think 1469 files have changed.. Could you maybe try re-submitting the request and see if it cleans things up a bit?

gbevin:
Ok, doing so

gbevin:
@julianstorer resubmitting didn't clean up anything, so I created a new fork and basically replaced the ProJucer files, committing them all in one go. Create this pull request: https://github.com/WeAreROLI/JUCE/pull/197

PR #154, opened 2017-01-23
Acurrate frame value

yhkk2585:
When I use pro tools, I have to change the value of 23.976 frame to
accurate value, so I recommend changing the value to 23.97602397


julianstorer:
Sorry, but we can't see what you're actually trying to show here, because of all the merge noise..

PR #155, opened 2017-01-30
Disable adding internal filters twice

harry-g:
Disable adding same internal filters twice, as all internal filters can be configured once and to one I/O only.
So imho it makes no sense adding them twice.
A transparent and intuitive solution to the user is disabling the already added filters in the menus.
Added a new method `hasFilter` to achieve this.


julianstorer:
Seems like a good idea, I'll add something to that effect..

PR #158, opened 2017-02-14
Fixed assert in ModuleDescription::getHeader() when moduleFolder is i…

ffAudio:
In ModuleDescription::getHeader() accessing a child file of the invalid moduleFolder triggers an assert.
It can be observed, when a project is opened and the location of modules has changed.


ffAudio:
Is merged to develop [here](https://github.com/julianstorer/JUCE/commit/c563d840da65d54c708b6a6b1950f80282180cdb) by tpoole. Thanks again

PR #159, opened 2017-02-21
Allow long running javascript to be stopped

gbevin:
This is helpful when a script is still running, but the application is shutting down.


julianstorer:
I can implement a stop() method much more efficiently by just making it timeout immediately... Is there a reason why you added a separate flag rather than doing that? Do you have a use-case where you need a different exception type to be thrown if it's interrupted?

gbevin:
It's mostly that the message printed to the console looks wrong then. Ie. the script didn't time out, it was purposefully stopped, but that's a detail. I'm not checking the string message of the exception to differentiate.

julianstorer:
OK, I've pushed something that should do the job.

gbevin:
Thanks Jules! :-)

PR #160, opened 2017-02-24
Add getCpuModel function and getCpuSpeedInMegaherz on Android

hm1992:
1. JUCE learned to get CPU brand string (e.g. Intel(R) Xeon(R) CPU E5-1650 v2 @ 3.50GHz) on OSX, Windows and Linux

This was tested on OSX 10.12, Windows 10 and CentOS 7

2. JUCE learned to get CPU type from "Hardware:" field in /proc/cpuinfo on Android

This was tested on OnePlus 3T

3. JUCE learned to retrieve the CPU speed on Android from a /sys/devices file

This was tested on OnePlus 3T. OnePlus 3T has 4 cores with 2 different clock speeds so it chooses the highest clock speed of the avaialble cores.


julianstorer:
Thanks Harry - that should end up on the develop branch soon!

PR #168, opened 2017-03-13
WavPack decoding support

c41x:
As in title.

### DIFF
--- a/modules/juce_audio_formats/codecs/juce_WavPackAudioFormat.cpp
+++ b/modules/juce_audio_formats/codecs/juce_WavPackAudioFormat.cpp
@@ -1,0 +1,315 @@
+/*
+  ==============================================================================
+
+   This file is part of the JUCE library.
+   Copyright (c) 2015 - ROLI Ltd.
+
+   Permission is granted to use this software under the terms of either:
+   a) the GPL v2 (or any later version)
+   b) the Affero GPL v3
+
+   Details of these licenses can be found at: www.gnu.org/licenses
+
+   JUCE is distributed in the hope that it will be useful, but WITHOUT ANY
+   WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+   A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+
+   ------------------------------------------------------------------------------
+
+   To release a closed-source product which uses JUCE, commercial licenses are
+   available: visit www.juce.com for more information.
+
+  ==============================================================================
+*/
+
+#if JUCE_USE_WAVPACK
+
+namespace WavPackNamespace
+{
+#include <wavpack/wavpack.h>
+}
+
+//==============================================================================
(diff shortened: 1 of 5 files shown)


modules/juce_audio_formats/codecs/juce_WavPackAudioFormat.cpp
```diff
@@ -0,0 +1,315 @@
...
+   available: visit www.juce.com for more information.
+
+  ==============================================================================
+*/
+
+#if JUCE_USE_WAVPACK
+
+namespace WavPackNamespace
+{
+#include <wavpack/wavpack.h>
```
jrlanglois: I think in true JUCE fashion wavpack's files would be added locally to the module.

julianstorer: Yes, we don't generally include 3rd party headers unless really necessary. What is in wavpack.h and where would someone get it? Does it rely on a static library to go with it, or is it header only?

jrlanglois: Looks like you can directly get the source: http://www.wavpack.com/downloads.html#sources

Alternatively, you could add a git submodule for its (official?) repository: https://github.com/dbry/WavPack

c41x:
I've included Wavpack source, only needed files, fixed warnings, and conversion errors.

jrlanglois:
Was curious, so ran the code on VS2013 and VS2015 - no warnings, which is good! Problem is that I can't find any files to test with. I guess this format is a bit niche?

c41x:
You're right, it is a bit niche (actually I am writing music player in JUCE so I want to support all formats). I think the easiest way to test it is to download converter (there is one in wavpacks site) and convert some audio to wv.

jrlanglois:
Thanks. I got a file converted using all sorts of different WavPack settings and they all played back just fine.

jrlanglois:
@c41x I'd be keen on using this in our application for similar reasons. Have you mentioned this on the forums?

c41x:
@jrlanglois Nope, I'm not registered on JUCE forums... If you write audio player you could use my FFmpeg decoder for JUCE (https://github.com/c41x/JUCE-FFmpeg) :)

jrlanglois:
@c41x I suggest you post that FFMPEG reader support on the forums - I think it will get a decent amount of attention!

jrlanglois:
@c41x I'm not so sure about the wavpack reader code after testing this with [wavpack's provided test suite](http://www.rarewares.org/wavpack/test_suite.zip).

jrlanglois:
Also, the sample rates supported by WavPack are as follows:
```c++
6000, 8000, 9600, 11025, 12000, 16000, 22050, 24000,
32000, 44100, 48000, 64000, 88200, 96000, 192000
```

You've only listed a subset:
```c++
8000, 11025, 12000, 16000, 22050, 32000,
44100, 48000, 88200, 96000, 176400, 192000
```

PR #175, opened 2017-03-27
Patch/url timeout

francoisbecker:


julianstorer:
OK, I see what you're trying to do, but this really doesn't fit for several reasons.. The URL class is a by-value class, so can't have mutating methods, and would need to copy this new member variable.

It could have a withTimeout() method, but at a semantic level, this really goes beyond the responsibility of the URL class, since a timeout isn't a property that goes with the address, it's only of interest to code that's responsible for opening the URL. Not sure what to suggest as a way to get around your problem, but this isn't the place for it. Maybe we just need to add options to other code that opens a URL.

francoisbecker:
Ok, yes, this doesn't fit very well. It's the smallest workaround patch I could devise. Or maybe use an options struct as a call parameter (including most of the parameters of the URL::createInputStream method)?

yfede:
> maybe use an options struct as a call parameter (including most of the parameters of the URL::createInputStream method)?

That would be the cleanest solution in my opinion too, I'm already kinda doing that myself for connections to web APIs and it works quite well

PR #176, opened 2017-03-31
Extended AudioPlayHead::FrameRateType to support a 60 fps flag.

jrlanglois:
The intent is to allow a host to specify a modern frame-rate, which VST2 and VST3 support.


julianstorer:
Cheers Joel, I'll sort that out for you today!

PR #180, opened 2017-04-18
Fixed import based on wrong module availability

gbevin:


julianstorer:
There seem to be no files changed in this pull request.. (??) Does that mean it's OK to close it?

gbevin:
I think it got probably fixed with JUCE 5's changes

PR #181, opened 2017-04-24
Drawable's copy constructor Fix

dowail:
Once **Drawable** is copied, two attributes related to **Component** gets reset to its default state. 
`setInterceptsMouseClicks()` and `setPaintingisUnclipped(), `are now set appropriately.


julianstorer:
Thanks, I'll get this sorted out today!

PR #183, opened 2017-04-26
Added comparison operators to Uuid following standard IETF and ITU conventions.

jrlanglois:
Sources and inspiration:

- https://www.ietf.org/rfc/rfc4122.txt
- http://oid-info.com/get/2.25
- http://www.boost.org/doc/libs/1_61_0/libs/uuid/uuid.html
- http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/util/UUID.java

Forum post: https://forum.juce.com/t/uuid-comparison-operator-improvements/22005

### DIFF
--- a/modules/juce_core/misc/juce_Uuid.cpp
+++ b/modules/juce_core/misc/juce_Uuid.cpp
@@ -53,8 +53,38 @@ Uuid& Uuid::operator= (const Uuid& other) noexcept
     return *this;
 }
 
+int Uuid::getTimeLow (const Uuid& uuid) noexcept                    { return ((int) uuid.uuid[15] << 24) + ((int) uuid.uuid[14] << 16) + ((int) uuid.uuid[13] << 8) + (int) uuid.uuid[12]; }
+int16 Uuid::getTimeMid (const Uuid& uuid) noexcept                  { return ((int16) uuid.uuid[11] << 8) + (int16) uuid.uuid[10]; }
+int16 Uuid::getTimeHiAndVersion (const Uuid& uuid) noexcept         { return ((int16) uuid.uuid[9] << 8) + (int16) uuid.uuid[8]; }
+uint8 Uuid::getClockSeqHiAndReserved (const Uuid& uuid) noexcept    { return uuid.uuid[9]; }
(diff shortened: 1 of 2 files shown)


julianstorer:
I think there are a few mistakes in there, and it all seems a bit over-complicated.. How about these as accessor methods for the UUID sections:

        /** Returns the time-low section of the UUID. */
    uint32 getTimeLow() const noexcept                  { return ByteOrder::bigEndianInt (uuid + 12); }
    /** Returns the time-mid section of the UUID. */
    uint16 getTimeMid() const noexcept                  { return ByteOrder::bigEndianShort (uuid + 10); }
    /** Returns the time-high-and-version section of the UUID. */
    uint16 getTimeHighAndVersion() const noexcept       { return ByteOrder::bigEndianShort (uuid + 8); }
    /** Returns the clock-seq-and-reserved section of the UUID. */
    uint8  getClockSeqAndReserved() const noexcept      { return uuid[7]; }
    /** Returns the clock-seq-low section of the UUID. */
    uint8  getClockSeqLow() const noexcept              { return uuid[6]; }
    /** Returns the node section of the UUID. */
    uint64 getNode() const noexcept                     { return (((uint64) ByteOrder::bigEndianShort (uuid + 4)) << 32) + ByteOrder::bigEndianInt (uuid); }

..and then a simple compare method that uses them?

jrlanglois:
Yep, as I said in the forum post this initial go at it is a bit goofy... Your suggestion works for me.

julianstorer:
Hmm, did you actually test any of your code, it just all looks the wrong way round.. Compare with the way that toDashedString() reads the bytes.

And since the whole thing is ordered as one big-endian number, shouldn't the compare method just be this?

int Uuid::compare (Uuid other) const noexcept
{
    for (int i = 0; i < 16; ++i)
        if (int diff = uuid[i] - other.uuid[i])
            return diff;

    return 0;
}

I'm curious as to what you were doing where your changes seemed to be a fix?

jrlanglois:
I guess I set myself up for the flac, but the code was basically copied from one of the sources listed above, and reformatted to JUCE's coding standard (to some extent...).

PR #187, opened 2017-04-28
Fix crash in Projucer's Android exporter

peterrudenko:
Steps to reproduce:
 - create gui app project,
 - set JUCE_PLUGINHOST_VST3 property enabled,
 - add Android exporter and try to select it


julianstorer:
Thanks, I'll push a fix for this shortly!

PR #198, opened 2017-05-03
Fix logic to determine the physical processor count on windows

cesaref:
The demo SystemInfo tab was showing the incorrect number of physical CPUs under Windows. The existing logic was failing, dropping through, and using the logical CPU count instead, so was only visible with hyperthreading enabled.

This now correctly reports the physical CPUs on my test box.


julianstorer:
Cool, thanks for the heads-up. I actually think a good fix might be to ignore this return value, as I somehow don't trust win32 to always be consistent in failing..

PR #205, opened 2017-05-23
Added saturation functionality to convertTo0to1

m-hilgendorf:
I'm constantly having to saturate the result of this function wherever I call it. This is a small change that just clips the result to the range [0.0, 1.0], without changing any existing functionality.

### DIFF
--- a/modules/juce_core/maths/juce_NormalisableRange.h
+++ b/modules/juce_core/maths/juce_NormalisableRange.h
@@ -138,25 +138,40 @@ public:
     /** Uses the properties of this mapping to convert a non-normalised value to
         its 0->1 representation.
     */
-    ValueType convertTo0to1 (ValueType v) const noexcept
+    ValueType convertTo0to1 (ValueType v, bool shouldSaturate = false) const noexcept
     {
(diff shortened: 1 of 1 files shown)


tpoole:
I'm afraid this isn't going to be included in JUCE - the whole point of a NormalisableRange is to provide a mapping between an arbitrary range and a normalised 0->1 range. If you're ending up outside of the normalised bounds then this sounds like a misuse of this class, where you're providing a non-normalised value outside of the previously specified range. Please submit a bug report if this isn't the case!

m-hilgendorf:
>where you're providing a non-normalised value outside of the previously specified range

Well that's exactly the use case I was experiencing. If you wind up with a value outside the normalized range, what should happen? At the moment the values are treated no differently, giving you negative values or values greater than one. If you rely on the system being "Normalised" that should mean that all values outside the range are clipped.

tpoole:
If you **expect** out of bounds values then you should clip them yourself. You could also provide your own lambda functions to the NormalisableRange constructor to do this for you.

A NormalisableRange should provide a bijective mapping between the two ranges - this is broken if multiple values map to zero or one.

PR #208, opened 2017-05-31
Fixed build error in VS2015 debug config

matok:
Error details:
juce_AudioProcessor.cpp(691): error C2059: syntax error: '}'

### DIFF
--- a/modules/juce_audio_processors/processors/juce_AudioProcessor.cpp
+++ b/modules/juce_audio_processors/processors/juce_AudioProcessor.cpp
@@ -688,7 +688,9 @@ void AudioProcessor::addParameter (AudioProcessorParameter* p)
     auto paramId = getParameterID (p->parameterIndex);
 
     for (auto q : managedParameters)
+    {
         jassert (q == nullptr || q == p || paramId != getParameterID (q->parameterIndex));
+    }
    #endif
 }
 


julianstorer:
Huh? How can it have a problem with a brace on line 691 when there isn't a brace there??

The original code looks completely fine to me, I can't see how you could possibly get that bug report from that code!

matok:
It looks like it's a bug in VS 2015 C++ compiler. (I have the latest update 3 installed)
In VS 2017 it works. I've also tried building it in Xcode using clang and it also works fine.

I didn't have much time to investigate it properly, I just tried to add braces and it worked.

If I'm not mistaken, the jassert macro on line 691 is being expanded to this line containing a few braces:
```C++
__pragma(warning(push)) __pragma(warning(disable:4127)) do { if (!(q == nullptr || q == p || paramId != getParameterID(q->parameterIndex))) __pragma(warning(push)) __pragma(warning(disable:4127)) do { juce::logAssertion(__FILE__, __LINE__);; if (juce::juce_isRunningUnderDebugger()) { __debugbreak(); }; } while (false) __pragma(warning(pop)); } while (false) __pragma(warning(pop));
```

You will get the same error also with this simple construction:
```C++
std::vector<int> a;
for (auto i : a)
    jassert(i);
```

while this works fine:
```C++
for (int i = 0; i < 10; ++i)
    jassert(i);
```

So it looks like it has something to do with the combination of iterator-based for loop and some of these braces in expanded `jassert` macro. But I suppose, if you wanted to make some workaround for this issue in VS 2015, it would be better to modify the macro, not just add braces here.. But I really don't know which one, maybe `JUCE_BLOCK_WITH_FORCED_SEMICOLON`? 

Maybe it's time to update my CI build machine to VS 2017... :)

julianstorer:
Wow, OK then! Thanks for the heads-up, we'll add some braces.

Actually.. I remember now hitting the same compiler bug myself and having to add some braces. Don't think there was a macro involved that time but I'm not sure exactly what it is that VS didn't like. Anyway, I'll sort this out!

PR #209, opened 2017-06-08
Suggestion: Replace global vars File::separator and File::separatorString with getters

matok:
This is just an idea, but I think it might be better to use static getters than global static vars in this case to avoid nasty order-of-initialisation problems..

I know it's not a good idea to call methods like these in static constructors but sometimes it can be handy, for example during debugging dynamic library loading (plugins for exotic/badly documented hosts or so)

Example - modification of JuceVersionPrinter that prints current executable path when library is being loaded - even if no entry point is called:

```C++
 struct JuceVersionPrinter
 {
     JuceVersionPrinter()
     {
         DBG (SystemStats::getJUCEVersion());
         DBG ("Loading from: " << File::getSpecialLocation (File::SpecialLocationType::currentExecutableFile).getFullPathName());
     }
 };

 static JuceVersionPrinter juceVersionPrinter;
```

### DIFF
--- a/modules/juce_core/files/juce_File.cpp
+++ b/modules/juce_core/files/juce_File.cpp
@@ -76,7 +76,7 @@ static String removeEllipsis (const String& path)
    #endif
     {
         StringArray toks;
-        toks.addTokens (path, File::separatorString, StringRef());
+        toks.addTokens (path, File::getSeparatorString(), StringRef());
         bool anythingChanged = false;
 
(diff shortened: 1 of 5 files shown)


julianstorer:
Yes, that would certainly be a better design. If only VS studio had better constexpr support, we could use that, but I might actually change this like you suggest. Will close this and push something today, thanks!

PR #224, opened 2017-07-20
Changes to Projucer Component Editor and Graphics element editor in o…

nonchai:
…rder to add the following functionality focused around setting width, height , x and y coordinates to more than one selected object at one time.

The operations added are:

** Align Left, Right, Top Bottom,  achieved by means of four new menu items added to the contextual menu when more than one object is highlighted.

** setting the width, height , x and y coordinates to more than one highlighted object by means of the properties inspector at the right hand side.

Whenever more than one object is selected the list of editable properties shown reduces to just those shared by all the highlighted objects - namely x,y, height and width.

All the source files changed were in subdirectories of Projucer/ComponentEditor.

Many methods have had an extra argument added to achieve this : bool multipleSelected.

In addition I had to move the enum ComponentPositionDimension into a different header file  - namely jucer_ComponentLayout.h

### DIFF
--- a/extras/Projucer/Source/ComponentEditor/components/jucer_ButtonHandler.h
+++ b/extras/Projucer/Source/ComponentEditor/components/jucer_ButtonHandler.h
@@ -36,9 +36,12 @@ public:
                                 defaultWidth_, defaultHeight_)
     {}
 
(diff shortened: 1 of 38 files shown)


JosephTLyons:
Is this not going to be included?

nonchai:
yes. just sorting out final issues before submitting another P.R.

PR #226, opened 2017-07-27
List Info Chunk original spec compliance

fredmerging:
Small change. Added extra fields to List Info Chunk to include original and official ICMT (comment) and IPRD (product or album name), and unoffical IPRT (another track number) but used by radios.
This solves a compatibility issue, original LIST Info Chunk spec (1991) and initial implementation of it i Juce did include ICMT and IPRD fields. And at the moment, there is no way to specify album name without IPRD.


hogliux:
This will appear on develop in a few minutes. Thanks for the suggestion.

fredmerging:
Thanks a lot!

PR #231, opened 2017-08-08
corrected obsolete function name

olilarkin:
prevented blocks drawing example from working


ed95:
The setLED() and blendLED() methods are needed for older firmware versions. The DrumPadGridProgram::getLittleFootProgram() method checks the firmware version of the connected block and returns a littlefoot program using either setLED() for versions pre-0.2.5 and fillPixel() for later versions.

PR #232, opened 2017-08-08
Proposed fix for compile errors in copy constructors

m-hilgendorf:
Fixes 'implicitly deleted copy constructor' compile error.


m-hilgendorf:
This was in Xcode 8.1, and it wasn't a warning. Any call to the copy constructors would fail to compile. 

That would be a nice warning to add in addition to the C++14 requirement in the projucer.

PR #234, opened 2017-08-11
Ignore some Autotools files

monocasual:
Ignore `.deps/` folder and `.dirstamp` file, two items generated by Autotools on Linux.


julianstorer:
Thanks - seems sensible, I've added that to develop now.

PR #235, opened 2017-08-13
Use constexpr for juce_Point.h constructors and methods when available

rsjaffe:
See https://forum.juce.com/t/constexpr-for-point-constructors-and-many-of-its-methods/22565/2 for discussion of this. I used the macro developed by Microsoft to detect earlier versions of Visual Studio to disable the constexpr keyword when necessary. The Microsoft code is under the MIT license, so merging it here won't cause license issues. See https://github.com/Microsoft/GSL/blob/master/include/gsl/gsl for the source of the macro.

### DIFF
--- a/modules/juce_graphics/geometry/juce_Point.h
+++ b/modules/juce_graphics/geometry/juce_Point.h
@@ -25,6 +25,10 @@
 */
 
 #pragma once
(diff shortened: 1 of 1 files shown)


hogliux:
Thank you for the pull request. This has been resolved with commit 4a2a529.

PR #239, opened 2017-08-15
Finish constexpr for Point.h

rsjaffe:
My apologies. Missed some functions that can be made constexpr with first pull.


modules/juce_graphics/geometry/juce_Point.h
```diff
@@ -25,6 +25,10 @@
 */
 
 #pragma once
+#if defined(_MSC_VER) && _MSC_VER < 1910
+#pragma push_macro("constexpr")
+#define constexpr /*constexpr*/
```
jrlanglois: I was under the impression that VS2013 has partial support for `constexpr`, so this code should be throwing errors at worst, warnings at best, no?

modules/juce_graphics/geometry/juce_Point.h
```diff
@@ -41,32 +45,32 @@ class Point
 {
 public:
     /** Creates a point at the origin */
-    Point() noexcept : x(), y() {}
+    constexpr Point() noexcept : x(), y() {}
```
jrlanglois: The JUCE team already added a series of changes like this on `develop`: https://github.com/WeAreROLI/JUCE/commit/4a2a5294a7bca077bedf9f68e4774075ac8843f8

rsjaffe: The macro erases constexpr keyword if using VS2013 or 2015. That's why it's there. Microsoft uses the macro themselves for this purpose.

rsjaffe:
Resolved conflicts with the way Juce is handling constexpr using a macro.

julianstorer:
Thanks! I'll sort this out and push something shortly.

PR #252, opened 2017-08-26
Added maximum number of input samples to LagrangeInterpolator

ffAudio:
This PR adds a maximum number of samples to the interpolate call. It also allows to use the interpolator to read from a circular buffer:

    AudioBuffer<float> outputBuffer (1, 1024);
    AudioBuffer<float> delayBuffer (1, 4.0 * sampleRate);
    int readPosition;
    int used = interpolator.process (speedRatio, 
                                     delayBuffer.getReadPointer (channel, readPosition),
                                     outputBuffer.getWritePointer (channel),
                                     outputBuffer.getNumSamples(),                /* numOut */
                                     delayBuffer.getNumSamples() - readPosition,  /* available */
                                     delayBuffer.getNumSamples());                /* rewind */
    readPosition = (readPosition + used) % delayBuffer.getNumSamples();

This also allows a safe usage, because an unfortunate high number for speedRatio can easily push the interpolator into unallocated memory for reading.

See https://forum.juce.com/t/circular-buffer-and-lagrangeinterpolator/23723/10 for background info and use-case.


julianstorer:
I think this extra functionality would need to be put into a separate method, or templated or use some other mechanism to avoid the overhead of adding the extra work to this method. 
The reason is that this can be a very sensitive performance hotspot in some situations (certainly in Tracktion when playing many resampled tracks it becomes the biggest bottleneck in the whole audio engine) so must be as optimal as possible.

ffAudio:
Yes, I fully agree. I think the maintenance cost for two separate versions is less than the performance penalty for checking on each sample, when it is not needed.
Shall I rewrite the PR or would you take over from here, as you would probably rewrite my code anyway, and I am sure you will come up with a more elegant solution...
Cheers!

julianstorer:
I don't have a test case for it, so wouldn't have time to implement myself, would be better if you had a go and then I reviewed it.

ffAudio:
Hi Jules, / @julianstorer 
I rewrote this PR so the faster unchecked process methods are still available. 
To make it clear for the user, I renamed the original versions to processUnchecked() and processAddingUnchecked().
The new variations differ by adding the arguments "available" (input samples to come) and a "wrapAround" (if end is reached, start from n samples before - circular buffer).
For backward compatibility I added the original methods calling processUnchecked() and left a deprecated note in the documentation.

Here is a use-case: https://github.com/ffAudio/ff_buffers/blob/master/ff_buffers_ElasticDelayBuffer.h
Just replace CircularLagrangeInterpolator with juce::LagrangeInterpolator, so it will work with that PR.

Hope you like it, it works here now in my CrazyDelay...

Cheers,
Daniel

julianstorer:
FYI I've put a refactored version of this on develop - let me know if you find any problems with it..

ffAudio:
Great, thank you! Had no problems, and having you looked over it makes me more confident :-)
Cheers!

PR #255, opened 2017-08-29
Added detection of Cubase 8.5 and 9

yfede:
(Cubase 8.5 is marketed as a separate product, exactly as Cubase 8 and Cubase 9)

This also takes care of preventing that:

`hostFilename.containsIgnoreCase ("Cubase 8")`

mistakenly classifies "Cubase 8.5" as "Cubase 8" instead

As for Cubase 9, it seems that starting with this version the scanning of plug-ins happens using an external executable named “vst2xscanner”, which is located inside the “Cubase 9.app” bundle on macOS, and inside the installation directory “C:\Program Files\Steinberg\Cubase 9” on Windows.

These changes take into account that, and also detects that scanner as Cubase 9 because it wouldn’t make sense to have a different behavior if you are running in the scanner rather than into the “real” Cubase 9.

### DIFF
--- a/modules/juce_audio_plugin_client/utility/juce_PluginHostType.h
+++ b/modules/juce_audio_plugin_client/utility/juce_PluginHostType.h
@@ -66,6 +66,7 @@ public:
         SteinbergCubase7,
         SteinbergCubase8,
         SteinbergCubase8_5,
+        SteinbergCubase9,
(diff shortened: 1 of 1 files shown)


tpoole:
Thank you for this, I'll get it added shortly.

PR #257, opened 2017-09-01
Fix for getting VST2 version codes right

yfede:
Explaination of the rationale behind this PR is found in the following post in the JUCE forum: 

https://forum.juce.com/t/issues-with-version-integer-reported-by-vst2/23867/16


yfede:
This has been properly integrated in JUCE with commit:
c38282778d060fac39ae7bf64905933d55866aaf

PR #264, opened 2017-09-23
Correct order of initialisation

h3xl3r:
If members are not declared in this order, use of the constructor 
```
MemoryOutputStream::MemoryOutputStream (const size_t initialSize)
  : blockToUse (&internalBlock), externalData (nullptr),
    position (0), size (0), availableSize (0)
{
    internalBlock.setSize (initialSize, false);
}
```

will lead to an allocation error, as the address of a not yet allocated object (internalBlock) is used in
```
blockToUse (&internalBlock)
```

### DIFF
--- a/modules/juce_core/streams/juce_MemoryOutputStream.h
+++ b/modules/juce_core/streams/juce_MemoryOutputStream.h
@@ -114,8 +114,8 @@ public:
 
 private:
     //==============================================================================
+    MemoryBlock internalBlock;    
     MemoryBlock* const blockToUse;
-    MemoryBlock internalBlock;
     void* externalData;
     size_t position, size, availableSize;
 


julianstorer:
Thanks, but you've misunderstood the code. It's just taking the address of that member variable. It doesn't matter whether or not it has been initialised at that point.

h3xl3r:
Don't think I am "misunderstanding" this (macOS 10.12.6, Xcode 9.0):
```
malloc: *** error for object 0x104644460: pointer being realloc'd was not allocated
*** set a breakpoint in malloc_error_break to debug
```

```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fffe5811d42 __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fffe58ff457 pthread_kill + 90
2   libsystem_c.dylib             	0x00007fffe5777420 abort + 129
3   libsystem_malloc.dylib        	0x00007fffe58680ad realloc + 413
4   net.hexler.vTEST              	0x000000010380d3b8 juce::HeapBlock<char, true>::realloc(unsigned long, unsigned long) + 88 (juce_HeapBlock.h:259)
5   net.hexler.vTEST              	0x00000001037f1374 juce::MemoryBlock::setSize(unsigned long, bool) + 116 (juce_MemoryBlock.cpp:131)
6   net.hexler.vTEST              	0x000000010380d4c2 juce::MemoryBlock::ensureSize(unsigned long, bool) + 66 (juce_MemoryBlock.cpp:154)
7   net.hexler.vTEST              	0x0000000103817045 juce::MemoryOutputStream::prepareToWrite(unsigned long) + 229 (juce_MemoryOutputStream.cpp:89)
8   net.hexler.vTEST              	0x0000000103817150 juce::MemoryOutputStream::write(void const*, unsigned long) + 128 (juce_MemoryOutputStream.cpp:112)
```

The test case is using the above mentioned constructor with default arguments and then writing to the stream.

Think about it, it is taking the address of ```internalBlock```, which at that point hasn't been created yet according to the rules of initialization of C++ (i.e. the order of declaration of the header file.)

I will have to keep the change as otherwise it's a plain crash for me and I can't use the class.

julianstorer:
No, honestly, you're wrong about this, and I don't appreciate you giving me a thumbs-down when I know my reply was correct, even if you did then edit your reply to tone down it down!

Like I said: it takes the ADDRESS of that member variable. It makes no difference whether or not the member has been initialised at that point - even if it contains garbage, it's address is perfectly valid, and all that happens is that the address gets stored, not used.

The crash you're seeing is *definitely* not caused by this order of initialisation, and your change is absolutely definitely not a fix - please don't do as you suggested and leave it in your code because it's masking your real bug, and it's only by pure chance that the memory corruption you're seeing doesn't happen to cause a crash when you change the memory layout of this class!

BTW the reason it's done that way is a minor optimisation to put the most-used pointer in the class at the class's own base address, which avoids the compiler generating a pointer offset each time it gets used. Maybe a trivial performance gain, but I can assure you that I know what I'm doing and the code would be no more "safe" if it was swapped over.

We use MemoryOutputStream in thousands of places in all our apps, and it gets very thoroughly unit-tested, so I'm confident is saying it's something else in your code that's messing up the stack. Probably a dangling pointer or something involving threads. I'm going to close this issue, but if you're stuck trying to track down the crash, maybe post on the forum some code which reproduces it, and we'll try to help spot what's wrong. (But most likely what'll happen is that in trying to create some test code to reproduce it, you'll discover your problem)

h3xl3r:
Yes, you are right. I did find the problem and it was external to MemoryOutputStream. Funny how the proposed change did reliably stop the funky behaviour though. Apologies for wasting both our time, mea culpa.

julianstorer:
Yeah, just changing memory layout can have dramatic effects. Glad you figured it out.

PR #276, opened 2017-10-01
Fixes to Filter copy constructors

m-hilgendorf:
Closed the old PR because this is more comprehensive. 

The copy constructors of FIR::Filter and IIR::Filter will cause compilation failure because they internally use the HeapBlock class, which has an explicitly deleted copy constructor. This fix works by copying the coefficients from the original instance and resets the filter to zero-state. 

I changed StateVariableFilter along with the others so the copy behavior resets the filter to zero state, which to me makes more sense than copying the filter state to a new instance which could be dropped on a different channel than the original.


julianstorer:
Thanks - I'll close this pull request and will add those methods in the juce code style.

PR #277, opened 2017-10-02
Added initial support for reading ID3 and Vorbis Comments.

jrlanglois:


kf6kjg:
Bumping this both here and there.  Being able to read the tags would be a very nice feature and it looks like this might have been lost in the woodwork.

PR #278, opened 2017-10-04
Fix 'pointer not be null' typo

0joshuaolson1:


ed95:
Thanks, this is on develop now with commit [93ed8ed](https://github.com/WeAreROLI/JUCE/commit/93ed8ed0fc54c9b924dbcad13b9aa2f4549e3c67).

PR #281, opened 2017-10-06
Added support for the Ogg Vorbis subformat in WavAudioFormat.

jrlanglois:


julianstorer:
I can't see how the ogg format manages to parse the stream if it looks like a WAV stream? Are you relying on the stream position being at the right point inside the WAV file when you pass it to the ogg vorbis reader?

..and where would someone find a suitable test file for this?

jrlanglois:
@julianstorer From my tests, the ogg reader doesn't seem to care where the stream is at in order to parse it. Looking back, that's probably a happy path with the approach and files I've been testing with. It might be wise to store the stream's original position and set it back to that before attempting to create an ogg reader - will update the PR.

Test files:
* [DL Breaker](http://www.jrlanglois.com/wavogg/DL%20Breaker.wav)
* [DL Broken](http://www.jrlanglois.com/wavogg/DL%20Broken.wav)

FYI I mentioned on the forums that I can't upload audio files there.

julianstorer:
Thanks! Will be on develop shortly..

PR #287, opened 2017-10-18
fix out of memory when do HTTP request excessively

airgiser:
excessive allocation of local references may cause the VM to run out of memory


julianstorer:
Thanks - will be fixed shortly!

PR #293, opened 2017-10-31
Add capability for Value to be updated synchronously

jamiebullock:
This is a non-breaking change that adds an additional parameter to getPropertyAsValue() allowing any Value::Listeners attached to the returned Value to be updated synchronously if desired.


julianstorer:
Thanks, looks like a good request, I'll take a look and merge something in shortly!

PR #295, opened 2017-10-31
Fix case to match the actual folder name

McMartin:
In 54029e776d, the case of "gradle" was changed to "Gradle" in `Projucer.jucer`, but not on disk.

@hogliux

### DIFF
--- a/extras/Projucer/Projucer.jucer
+++ b/extras/Projucer/Projucer.jucer
@@ -155,12 +155,12 @@
             file="Source/Application/jucer_MainWindow.h"/>
     </GROUP>
     <GROUP id="{8853BD6C-8307-BCF8-65BE-6A2F57093B40}" name="BinaryData">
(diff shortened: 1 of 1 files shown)


McMartin:
Due to this bug, `Projucer` 5.2.0 cannot be built with [FRUT] (https://github.com/McMartin/FRUT) on Linux (which is case-sensitive) and I had to exclude `Projucer` 5.2.0 from my CI  (see https://github.com/McMartin/FRUT/commit/bf73515ed7287f9779f697f7a7c738dd691dce82). I haven't tried to save `Projucer.jucer` on Linux, but I guess it will also fail. 
@hogliux @julianstorer

ed95:
Sorry for the delay on this. Fixed [here](https://github.com/WeAreROLI/JUCE/commit/9bb5c2e32ea8624f1caa600a20e1d3bf2442e420)

McMartin:
Thanks!

PR #300, opened 2017-11-07
Projucer: Fix for null-terminated strings error with BinaryData::namedResourceList[]

COx2:
Currently, generated BinaryData::namedResourceList[] property is not ended with nullptr.
It causes the null-terminated strings error in the case like below. 

 `StringArray(BinaryData::namedResourceList) `

I made a change to solve this inconvenience.
If there is no reason for the current implementation, I hope that you will accept this request

### DIFF
--- a/extras/Projucer/Source/ProjectSaving/jucer_ResourceFile.cpp
+++ b/extras/Projucer/Source/ProjectSaving/jucer_ResourceFile.cpp
@@ -239,8 +239,9 @@ Result ResourceFile::writeCpp (MemoryOutputStream& cpp, const File& headerFile, 
             << "{" << newLine;
 
         for (int j = 0; j < files.size(); ++j)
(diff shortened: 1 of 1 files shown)


julianstorer:
It's wrong to describe this as a "fix" - the BinaryData class does exactly what its documentation describes - it doesn't claim that the list will end with a nullptr! 

The error isn't in BinaryData, it's in the fact that you're calling the wrong StringArray constructor -
 it has another one that takes an array + size, and that's the one you should use.

TBH if I was going to change anything here, then adding some begin()/end() iterators to the BinaryData class would make more sense, so standard algorithms can iterate it.

COx2:
@julianstorer 
Thank you for your answering in detail.
Certainly, using a different StringArray constructor that takes an array + size, I could handle it without problem.
Thank you for your advice.

PR #303, opened 2017-11-10
VST3 SDK v3.6.8 update patch

francoisbecker:
Here is most of the work that is needed for supporting the new VST3 SDK.
The sole part that remains to fix is the one with a `// TODO` in `juce_VST3_Wrapper.cpp`: it runs but I suspect that it works by accident.


francoisbecker:
Actually the code is correct, the // TODO can be removed.

tpoole:
The VST3 SDK v3.6.8 is now supported on the `develop` branch.

PR #308, opened 2017-11-19
remove duplicate gridItemPanel

modosc:
having two panels with `0` in them makes it slightly confusing to parse what's happening in the demo.


ed95:
Thanks. Fixed [here](https://github.com/WeAreROLI/JUCE/commit/504b95b5a6352e73b176babbb83c21eb8fc8ba63).

PR #309, opened 2017-11-21
Remove member variables which are always empty

McMartin:
I found these empty/unused member variables when studying the Xcode exporter code for [FRUT](https://github.com/McMartin/FRUT).

@ed95 @hogliux @julianstorer @tpoole: merge at will!


tpoole:
Thanks, we'll sort this out.

PR #311, opened 2017-11-24
Make SparseSet<>::operator==/!= const

pipou:
SparseSet<>::operator== and != should be const, as they only rely on the underlying Array comparison operators, which are const.


julianstorer:
Thanks for spotting that! Will push a correction shortly!

PR #318, opened 2017-12-11
Added previous/next window shortcuts for Projucer

ffAudio:
Added Ctrl+Tab shortcut to skip to next Projucer window. Also Ctrl+Shift+Tab raises the previous window (and moves focus there...)


julianstorer:
Thanks! Have pushed a version of this now

PR #319, opened 2017-12-12
Update juce_Analytics.h

johanneskares:
When using the analytics module in a Pro Tools plugin it always crashed. This fix should solve it.


julianstorer:
Ah, good catch, thanks! A fix should appear on develop shortly!

PR #327, opened 2017-12-22
Fix for matrix multiplication with different outer dimensions.

thomasdeppisch:
Just a short fix for the problem described in https://github.com/WeAreROLI/JUCE/issues/326 .


tpoole:
Yes, you're right. I'm not sure how that assertion got there...

We'll get this fixed - thanks for reporting.

PR #328, opened 2017-12-22
Added BSD environment

Martinfx:


modules/juce_core/system/juce_TargetPlatform.h
```diff
@@ -63,7 +63,7 @@
 #elif defined (JUCE_ANDROID)
   #undef        JUCE_ANDROID
   #define       JUCE_ANDROID 1
-#elif defined (LINUX) || defined (__linux__)
+#elif defined (linux) || defined (__linux__)
```
julianstorer: Are you sure about this? I've definitely seen the upper-case version being used!

Martinfx: I looked for on  page https://sourceforge.net/p/predef/wiki/OperatingSystems/ .

Martinfx: Where did you see upper-case version ?

Martinfx: and here: http://www.faqs.org/docs/Linux-HOWTO/GCC-HOWTO.html

julianstorer: OK, that sounds pretty convincing, but is there a reason to actually remove the upper-case one? I'd rather leave it in, just in case removing it would break people's builds.

julianstorer: (Can't remember where I saw it used.. this was years ago..)

Martinfx: Because this macro is badly.

julianstorer: I'm confused - why would a BSD build have that macro set?

Martinfx: I am sorry this is my mistake in my BSD build, but i think that macro does not exist

julianstorer: OK. Thanks for the other suggestions though, I'll add those BSD includes for you!

tpoole:
Added in https://github.com/WeAreROLI/JUCE/commit/7c847e7126b1fa55b07392c217a7b3d14da2d6af.

PR #329, opened 2017-12-25
Remove hardcoded label colours

peterrudenko:
Hi guys,

I'd very much appreciate if you remove that hard-coded colours from Label class:
```
-    setColour (TextEditor::textColourId, Colours::black);		
-    setColour (TextEditor::backgroundColourId, Colours::transparentBlack);		
-    setColour (TextEditor::outlineColourId, Colours::transparentBlack);
```

The reason is that they make Projucer's code-generation problematic: Projucer generates colour-setting code for components for every registered colour *that was set*, so once Label sets that colours by itself in a constructor, there's no way to use default colours for label's editor. And every GUI component that has labels on it always ends up having auto-generated code like this:
```
    trackNameEditor->setColour (TextEditor::backgroundColourId, Colour (0x00000000));
    trackNameEditor->setColour (TextEditor::outlineColourId, Colour (0x00000000));
```

which is not only useless and superfluous, but also messes up an existing colour scheme.

Steps to reproduce:
1. Add a label in a GUI component in Projucer (a generated code will have that colour setters)
2. Reset all label's colours with `Reset To Default` button
3. Save and make sure there are no superfluous colour setters
4. Restart Projucer to reload and resave the project
5. Now a generated code has that colour setters again


julianstorer:
Hmm, removing that will almost certainly break people's existing code, so I can't do that. 

If the problem is in the Projucer's GUI editor (which we're not actively developing anyway so doesn't justify a breaking change to Label) then probably a better fix would be to add a workaround to stop it generating code to set colours that are already there by default in a default-constructed instance of that class?

PR #330, opened 2017-12-30
Fixed component with abolsute position has its bounds set in resized()

dkollmann:
I encountered an issue that the bounds of a component was set in resized() instead of the constructor although it has an absolute position. When the anchor for a component is set to top/left (1) it would not considered to have an absolute position since the x/yMode has a value of 9 instead of absoluteFromParentTopLeft (8). So I basically ignore the top/left anchor when checking if the mode is absoluteFromParentTopLeft.

It would be great if someone could have a look at this fix who has more experience with the JUCE layouting code.

I did this change based on the 5.2.0 release code.


julianstorer:
Hmm, this may or may not be a valid correction - I'll have a quick look.. But more importantly, we're actively moving towards deprecation of these RelativePosition classes, so my advice would be to find a different way of doing your layouts before you waste too much time building a system based on this stuff!

julianstorer:
Oops, sorry - just took a look, and I'd misunderstood where this code was - didn't realise you were talking about code in the Projucer, and my comment about the relative position classes being deprecated is irrelevant! Thanks for the heads-up, I'll fix this now!

PR #336, opened 2018-01-08
VST3 Parameters added to receive MIDI CCs shouldn't be listed as automatable

yfede:
MIDI CCs should be the only way to control those parameters. What will
happen if they are marked as automatable, and the DAW controls them
both via automation and MIDI CCs? Sure cause of conflicts.

The host applications (SONAR, Cubase, REAPER to say just a few) don't
need these parameters to be marked as automatable at all for them to work.

Please take into account that we're delivering since years plug-ins in VST3 format that receive CCs without problems with this patch applied.

Screenshot below is taken from Presonus Studio One 3

screen shot 2018-01-08 at 18 02 58


tpoole:
I think that change sounds sensible.

Thanks for reporting!

PR #340, opened 2018-01-13
[Docs] - JUCE/Readme.md - fixing minor issues with docs

tiagomoraismorgado88:
#### [Docs] - JUCE/Readme.md - fixing minor issues with docs


tpoole:
It's difficult to see exactly what you've changed because all of the new line lengths are different. However, having picked through the changes, it looks like there are only a couple of commas that have been taken out. These commas have come from our legal team, so we're going to leave them in.

PR #341, opened 2018-01-14
Listing MIDI inputs on Linux shows outputs.

mtytel:
Got a report that Linux MIDI outputs were appearing in the MIDI inputs list. They traced it back here.
https://github.com/mtytel/helm/issues/176
https://github.com/mtytel/helm/issues/177


tpoole:
Thank you for reporting - we'll get that fixed.

I'm rather surprised that the current behaviour has existed for so long!

mtytel:
Yeah I think this only appears in certain cases. I've never seen the bug myself.

PR #342, opened 2018-01-15
Fix and cleanup module dependencies

McMartin:
While working on CMake support inside JUCE, I looked at the header file of each JUCE module and I found some weird things in the module dependencies.

@julianstorer @tpoole


tpoole:
Fixed here: https://github.com/WeAreROLI/JUCE/commit/7d943e76dc17d0d75a5a9d61505dc7bfc13040da

PR #348, opened 2018-02-06
Add 'isIdentity' flag in TranslationOrTransform

pipou:
This flag allows to avoid unnecessary transform or data duplication in the software based renderer.
See commit comments.

### DIFF
--- a/modules/juce_graphics/native/juce_RenderingHelpers.h
+++ b/modules/juce_graphics/native/juce_RenderingHelpers.h
@@ -42,13 +42,13 @@ class TranslationOrTransform
 {
 public:
     TranslationOrTransform (Point<int> origin) noexcept
-        : offset (origin), isOnlyTranslated (true), isRotated (false)
+        : offset (origin), isIdentity (origin.isOrigin()), isOnlyTranslated (true), isRotated (false)
     {
     }
 
     TranslationOrTransform (const TranslationOrTransform& other) noexcept
         : complexTransform (other.complexTransform), offset (other.offset),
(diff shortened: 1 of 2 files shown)


julianstorer:
Thanks for spotting this as a potential speed-up! But looking at the changes, it looks like the only really expensive operation that it avoids are the two places where a RectangleList needs to be offset, so wouldn't it make more sense for those places to just check isOrigin() directly, rather than needing to maintain this extra isIdentity flag?

pipou:
Yes indeed, the main reason I did this change in the first place was to handle large RectangleLists, and I was write surprised by the perf improvement..
And you're right, in all the other places, the overhead of the extra transform is negligible.
I choosed the extra flag in order to be consistent with the isTranslationOnly and isRotated, and the isIdentity looked like the best approach, as it applies to either the offset or the complexTransform, so it can be tested no matter what the current transform type is, furthermore, it's quite easy to maintain it as there are only a couple of places that can change the nature of the transform. It looked like a better option than a isTranslated flag (which would have required an additional isScaled flag if you look deeper...), and I wanted to keep the almost methood-free design for the TranslationOrTransform helper class.
Having this extra flag is not mandatory, but it doesn't hurt and allows a finer granularity when handling the transform.
It's up to you at the end, as long as I can have the RectList optimization, I'm happy :)
Thanks for looking into this!

julianstorer:
The code will be much less branchy without this flag, so I'll do something like that. Cheers for spotting an optimisation opportunity!

PR #350, opened 2018-02-09
NS_AVAILABLE instead of MACROS (Features enabled/disabled on runtime instead of during compilation time)

BastienC09:
I suggest using NS_AVAILABLE instead of the macros : JUCE_AUV3_VIEW_CONFIG_SUPPORTED and  JUCE_AUV3_MIDI_OUTPUT_SUPPORTED

The reason is that it would allow to use MIDI OUTPUT and VIEW CONFIG without changing the Deployment Target to the minimum required for the features. The features would be enabled only if the version of the OS is correct.

You might need to add a getter to replace the MACROS to know if the features are available in case some people need this info in their app.

Thanks you

### DIFF
--- a/modules/juce_audio_devices/audio_io/juce_AudioIODevice.h
+++ b/modules/juce_audio_devices/audio_io/juce_AudioIODevice.h
@@ -111,6 +111,8 @@ public:
         this callback.
     */
     virtual void audioDeviceError (const String& errorMessage);
+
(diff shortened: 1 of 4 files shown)


BastienC09:
Part 2/ 
I moved "copy back" to the scope above as if this scope is not called there is nothing to copy back.
Cause app to crash at times
--
All this was tested on iOS only.

PR #351, opened 2018-02-10
Use JucePlugin_Name as (preferred) fallback jack-client name

umlaeute:
and use the literal "*JUCEJack*" only as a last resort.

this will make a plugin named `FooBar` appear as "*FooBar*" in jack (rather than as "*JUCEJack*")

Closes: https://github.com/WeAreROLI/JUCE/issues/331


ed95:
Added in [e5c3837](https://github.com/WeAreROLI/JUCE/commit/e5c3837f057e604d9f96485847ff0d43f49584ac) - thanks for the PR!

PR #356, opened 2018-02-27
Added access to AudioProcessorValueTreeState::Parameter's valueToText…

ffAudio:
…Function and propagate via SliderAttachment

* Exposes the lambdas in the AudioProcessorValueTreeState::Parameter
* Adds lambdas to the Slider to convert value to text and vice versa
* AudioProcessorValueTreeState::SliderAttachment propagagtes the lambdas to the slider on creation

A slider connected via an AudioProcessorValueTreeState::SliderAttachment will automatically use the lambdas of the connected Parameter to display the text and to convert it back


tpoole:
This is now available in 5.4.1.

PR #362, opened 2018-03-06
Added Line::getYFromX and Line::getXFromY

modosc:
also add Line::rangeContains and Line::domainContains methods.


ed95:
It's easy enough to add these as free functions in your own code and probably not something we'd add to the class directly in JUCE.

PR #370, opened 2018-03-31
Remove dead code in MakefileProjectExporter

McMartin:
Follow-up of #369.

@ed95 @hogliux please merge at will, thanks!


ed95:
Thanks, I'll get both of those merged in.

PR #378, opened 2018-04-13
remove 20 decimals forcing when converting from double to String

benkuper:
When converting from double to String, the default behavior should not be having 20 decimals but automatic decimal precision


julianstorer:
When changing the type of a variant it's vital for the precision to stay as high as possible - that's why it's deliberately setting the precision and not using the default (which is both too low and not necessarily platform-independent). 
And even if your change was an improvement, it still wouldn't be a change we'd want to make because it'd break existing code that relies on the current behaviour!

julianstorer:
Don't really understand what you mean by "official"? We can't provide built in methods for every possible way that people might want to use a double value, but it's trivial to get the double value and pass it to whatever conversion function is appropriate for your app.

PR #379, opened 2018-04-14
Allow ComponentEditor to edit source files other than .cpp

jpcima:
It is not permitted to edit GUI components unless the files are named .cpp.
I permit the other C++ file extensions.


ed95:
Thanks, I'll add this.

PR #399, opened 2018-06-25
Add OSC Color support

benkuper:
because int32 and uint32 are recognized as the same type, i had to add a bool flag to the OSCArgument(int32) constructor to choose whether its a color or an int

### DIFF
--- a/modules/juce_osc/osc/juce_OSCArgument.cpp
+++ b/modules/juce_osc/osc/juce_OSCArgument.cpp
@@ -27,11 +27,9 @@
 namespace juce
 {
 
-OSCArgument::OSCArgument (int32 value, bool asColor) noexcept
-    : type (asColor?OSCTypes::color:OSCTypes::int32)
+OSCArgument::OSCArgument (int32 value) noexcept
+    : type (OSCTypes::int32), intValue(value)
 {
-	if (asColor) colorValue = value;
-	else intValue = value;
 }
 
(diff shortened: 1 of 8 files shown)


benkuper:
@julianstorer I added an OSCColor struct as per your remarks on the other PR, is it what you had in mind ?

julianstorer:
Yeah, more or less, though I'd probably do a few things differently. Anyway, thanks for the heads-up, I'll have a go at implementing something like this soon, maybe today. :)

BTW, pro-tip: I don't know if you looked at the github diff for this PR, but mixing up tabs and spaces and not matching the style of adjacent code makes it look a real mess. 

We always refactor any code that might go into JUCE, so formatting isn't a big deal to us (except for making the diff hard to read), but if you're ever contributing to other projects where they directly merge from PRs, you'll look at lot more professional if you take the trouble to match their style.

benkuper:
Ok thanks !
Yeah i saw the indentation problem but what i don't understand is that i don't have this indentation in VS2017 (i'm actually also maniac on that), i'm using gitkraken but i don't know exactly where does it come from.
I'll try to have a better check at the output next time ;)

julianstorer:
Cheers. Yeah, I don't know what you did but I can't even apply the diff successfully to see your changes in my local tree - both the diff and patch fail because of trailing whitespace and other problems. Ah well, probably quicker to just re-write it!

PR #400, opened 2018-06-25
add Colour::getRGBA function

benkuper:


julianstorer:
(Thanks for pointing out the typo!)

I don't think this is a good addition though - if there are methods for returning more than one RGB format, then it leads to feature-creep as it looks like we should support all the other packing orders too, and that doesn't belong in this class. 

Also, I can't really think of a use-case where it would be better for us to add this method than for people to write their own one-line free function that converts a Colour to whatever format you're trying to create.

benkuper:
I can think of 2 reasons :)
First, there already is a Colour::fromRGBA method (and no fromARGB one, strangely ?), so that would make sense to do both ways here imho. 
Second, and it's linked to the other PR, the "official" color format in OSC messages is RGBA.
I understand that multiple practices are not good, it just already seem that both formats are there somehow (and ARGB & RGBA are the 2 main ones, at least outside of GPUs).

julianstorer:
Yes, I saw your other PR, which is a good subject for a request, but the way you did it using ints as colours isn't the right approach - given that you can't use juce::Colour in the OSC module, the right alternative isn't to fall back to using ints - instead if you create a small OSC::Colour struct that holds the RGB values, then you'd be using the type system rather than a bool (!) to distinguish between an int and a colour. A free function could easily convert between the two Colour classes. (Sorry, this comment probably belongs on your other PR!)

And no, it's missing the point to say "there's RGBA so where is ARGB?". A better question would be "why is there any method at all to return an int, it's not the responsibility of the Colour class!). 
Also, the Colour class has a constructor that takes an int, so if it had two methods that return different ints, it becomes completely unclear which one produces the right type of number for the constructor! That'd be totally unacceptable in an API.
It's my fault - I wrote that stuff many years ago and wouldn't do it like that now. A better way would have been to provide a more versatile packing system that can create an int with any order, perhaps using a templated class to hold colour components. And probably make it separate to the Colour class.

PR #416, opened 2018-09-04
Passes the function parameter 'areaToFitWithin' of the method constra…

Xav83:
…inedWithin by const reference.

Hi JUCE developpers,

I've create this pull request to help improving JUCE performance by avoiding a copy of `juce::Rectangle` in the method `constrainedWithin` in `juce_Rectangle.h` file.
The parameter was passed by value, so a copy of the parameter was made, but only `const` method are called on this object in this method. So passing the object by `const` reference will avoid this copy while assuring that the object is not modified.

If you have any recommendation, or suggestion to improve this PR, don't hesitate to ask.

Thanks for your work good work on JUCE.
Hope to ear about you soon,

Xavier Jouvenot

### DIFF
--- a/modules/juce_graphics/geometry/juce_Rectangle.h
+++ b/modules/juce_graphics/geometry/juce_Rectangle.h
@@ -767,7 +767,7 @@ public:
         is larger than the target rectangle in either dimension, then that dimension
         will be reduced to fit within the target.
     */
-    Rectangle constrainedWithin (Rectangle areaToFitWithin) const noexcept
+    Rectangle constrainedWithin (const Rectangle& areaToFitWithin) const noexcept
     {
         auto newW = jmin (w, areaToFitWithin.getWidth());
         auto newH = jmin (h, areaToFitWithin.getHeight());


tpoole:
It is a little counterintuitive, but passing small objects with primitive data members by value is more efficient than passing by const reference. By making a copy of an object the compiler can optimise more aggressively (including not actually doing the copy) as the function is much more constrained in what it can do. A const reference is essentially a pointer, which it is possible index off to modify program state in other locations.

Xav83:
Good to know 😃
Thanks for the information @tpoole 

Just for not making this mistake again, what is the limit of the object's size for them to be consider small enough by the compiler to optimise that way, if there is any rule about that ?

Instead of passing it by `const` reference, does passing the parameter as `const` allows the compiler to optimize it more easily, knowing that it won't be modified by the function ?
`Rectangle constrainedWithin (const Rectangle areaToFitWithin) const noexcept`

tpoole:
No hard rules about the size of the objects I'm afraid; it's somewhat context dependent so you need to profile to get firm answers.

`const` doesn't help the compiler either as a programmer is free to `const_cast` the qualifier away and modify the object anyway - `const` is more a tool to suggest to the programmer the semantics of how something is expecting to be used. In this particular case there's no logical reason why the function argument needs to remain unchanged when we return from the function, so there's no advantage to making it `const`.

Xav83:
Thanks for all your explanation @tpoole 😃

PR #418, opened 2018-09-16
Fixed the build on gcc 4.9 not implementing std::is_trivially_copyable

jpcima:
Gcc 4.9 does not have a complete c++11 implementation which includes `std::is_trivially_copyable`.
If I detect a gcc compiler version < 5, I disable the copy optimization which prevents the build on this older compiler.


tpoole:
Thanks for this!

I've just made some similar changes, and a few more, here d06a2b8c13ff123503282d34d59f9146e1093125.

PR #419, opened 2018-09-18
Add AVX512 feature detection

hjmallon:
I realised I would need to detect AVX512 so I have added some detection of it. I haven't done every AVX512 feature flag, but I have done most.


tpoole:
Added in bedff619838dd65c617d043c91cdf5bad4fe605d.

PR #426, opened 2018-10-09
fix typo in modules/juce_audio_devices/audio_io/juce_AudioDeviceManager.h

bill-auger:
the doc strings are identical for both AudioDeviceManager->getInputLevelGetter() and AudioDeviceManager->getOutputLevelGetter()


McMartin:
@bill-auger line 438 has the same typo: it should read `The current output level can be read`.

tpoole:
Thanks, will get that fixed.

PR #430, opened 2018-10-30
Added support for WebBrowserComponent::loadHTMLString()

jkoutavas:
A simple demo of this can be found at https://github.com/jkoutavas/juce_HTMLStringDemo

### DIFF
--- a/modules/juce_gui_extra/native/juce_win32_WebBrowserComponent.cpp
+++ b/modules/juce_gui_extra/native/juce_win32_WebBrowserComponent.cpp
@@ -24,6 +24,8 @@
   ==============================================================================
 */
 
+#include <mshtml.h>
+
 namespace juce
 {
 
@@ -44,7 +46,7 @@ JUCE_DECLARE_UUID_GETTER (WebBrowser,                "8856F961-340A-11D0-A96B-00
 class WebBrowserComponent::Pimpl   : public ActiveXControlComponent
 {
 public:
-    Pimpl() {}
+    Pimpl() : htmlSource(nullptr) {}
 
     ~Pimpl()
     {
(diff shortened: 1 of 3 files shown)


modules/juce_gui_extra/native/juce_win32_WebBrowserComponent.cpp
```diff
@@ -74,13 +76,68 @@ class WebBrowserComponent::Pimpl   : public ActiveXControlComponent
...
         }
     }
 
+    void loadHTMLString(const String& htmlString, const String& /*baseURLString*/)
+    {
+        if (htmlSource)
+        {
+            delete[] htmlSource;
+        }
+        htmlSource = new wchar_t[htmlString.length() + 1];
```
martinfinke: When `WebBrowserComponent::Pimpl` is destroyed, it seems like `htmlSource` will be leaked. Would it be possible to use `std::unique_ptr<wchar_t[]>` or `std::vector<wchar_t>` instead? This would also avoid double-delete if `loadHTMLSource` is called after `loadHTMLString`.

jkoutavas: Okay, I made htmlSource a juce::String and made the lifetime of `new wchar_t[htmlSource.length() + 1];` contained within `loadHTMLSource()`

martinfinke: The lifetime is still spread across two functions (loadHTMLString and loadHTMLSource). Why not just use std::vector<wchar_t>?

jkoutavas: Lifetime of what? With my latest commit, htmlSource is now a class member variable juce::String. The new "context" allocation/deallocation is happening all within the scope of loadHTMLSource()

PR #431, opened 2018-10-31
fix typo in modules/juce_events/messages/juce_ApplicationBase.h

bill-auger:
the typo is just the "but it you will" bit; but these could use the extra details and clarification that it is the running instance that gets the callback, and only on some platforms


ed95:
Added in [aae372d](https://github.com/WeAreROLI/JUCE/commit/aae372d68c4b6b01581fe3b3df522182d428dde8).

PR #436, opened 2018-11-13
Update Android project template to have recommended default values

dturner:
Make the default project template for Android conform to current Android developer guidelines. Changes as follows:

1) Change the gradle version to 4.10 and the plugin version to 3.2.1. These are the recommended settings for projects in the current stable release of Android Studio (also version 3.2.1). 

2) Remove `buildToolsVersion` as it's no longer required

3) Add a new variable for `targetSDK` and set it to the latest version of Android - 28 (Pie). Google Play requires new apps target API 26 or above: https://developer.android.com/distribute/best-practices/develop/target-sdk

4) Update the ABIs to support arm64-v8a and x86_64 (for the Android emulator). This required removing the `-DANDROID_PLATFORM` preprocessor define as this will be handled by the NDK's cmake configuration (which will use platform-16 for 32bit targets and platform-21 for 64bit targets).


lukaszkozakiewicz-roli:
Merged to develop.

PR #449, opened 2018-12-13
juce_AttributedString: copying attribute before insert

igfraga:
This avoids an assert and the possibility of a dangling reference.


igfraga:
Fixed in develop

PR #452, opened 2018-12-14
AAX:  Call setVisible (false) on the editor content component before …

tom-sonnox:
…removing it from the desktop to avoid an OpenGLTexture leak


ed95:
See [5c03d13](https://github.com/WeAreROLI/JUCE/commit/5c03d13f0dfcb06b19a77f37bd9c82a4022ea411).

PR #455, opened 2018-12-20
PROJUCER: Added per-file compiler settings

Jazzdoodle:
As discussed recently on the JUCE forum, this is an implementation of individual compiler settings for code files in Projucer. All exporters are supported, with exception of the incomplete CLion exporter.

Code files receive a new property called a compiler flag setting. New settings can be created and assigned in the file browser. The actual flag configurations belonging to each setting are then set in the exporter properties.

### DIFF
--- a/extras/Projucer/Source/Project/UI/jucer_FileGroupInformationComponent.h
+++ b/extras/Projucer/Source/Project/UI/jucer_FileGroupInformationComponent.h
@@ -37,8 +37,8 @@ public:
         : item (group),
           header (item.getName(), { getIcons().openFolder, Colours::transparentBlack })
     {
-        list.setHeaderComponent (new ListBoxHeader ( { "File", "Binary Resource", "Xcode Resource", "Compile" },
-                                                     { 0.4f, 0.2f, 0.2f, 0.2f } ));
+        list.setHeaderComponent (new ListBoxHeader ( { "File", "Binary Resource", "Xcode Resource", "Compile", "Compiler Flags Setting" },
+                                                     { 0.4f, 0.2f, 0.2f, 0.2f, 0.4f } ));
         list.setModel (this);
         list.setColour (ListBox::backgroundColourId, Colours::transparentBlack);
         addAndMakeVisible (list);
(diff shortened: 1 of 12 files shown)


McMartin:
I built and tested this version of Projucer, but it doesn't seem to do what I expected...

I wanted to compile a file with `/Wall` (the MSVC flag to enable all warnings including those that are disabled by default), so I clicked on `Add new setting ...` and typed `/Wall`. Then I pressed `ENTER` and `/Wall` became `Wall`, for whatever reason the `/` got deleted... Then I saved and I looked at the diff of the generated Visual Studio project:
```diff
@@ -138,7 +138,9 @@
     <Lib/>
   </ItemDefinitionGroup>
   <ItemGroup>
-    <ClCompile Include="..\..\Source\foo.cpp"/>
+    <ClCompile Include="..\..\Source\foo.cpp">
+      <AdditionalOptions> %(AdditionalOptions)</AdditionalOptions>
+    </ClCompile>
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="..\..\JuceLibraryCode\AppConfig.h"/>
```
What did I do wrong?

Another thing I'm wondering is how to deal with platform differences. I would like to compile my file `foo.cpp` with `-Weverything` when using clang (macOS and iOS), with `/Wall` when using MSVC (Windows) and with `-Wall -Wextra -Wpedantic ...` (and a bunch of others) when using gcc/mingw (Linux/ Windows). I doesn't seem to be possible currently.

Jazzdoodle:
HI,

 my pull request had a small explanation how it works. The settings you add in the file browser just define setting names. So in your case you would add a setting that is called "All_Warnings", for example. Then you go to the exporter property page and find a new property text field for "Extra Compiler Flags for Setting All_Warnings". There you can add the actual flag configuration for the compiler the exporter uses. There you can also set several flags, like -Wall -Wextra -Wpedantic.

That means every exporter can have different flag configurations, which it will in general need, because flags are not consistent between different compilers.

I hope that makes it clearer.

Thanks,

 Andreas

ed95:
This has now been added in [b2e2346](https://github.com/WeAreROLI/JUCE/commit/b2e2346745eeb0fcdf2263fd7fec2f318c27f567) - thanks for the PR and sorry for the delay!

PR #459, opened 2019-01-04
constexpr Range constructor

CrushedPixel:
I needed the two-argument constructor for `juce::Range` to be `constexpr`, so that I can create `static const` class members inline:

```cpp
class Notes {
public:
	static constexpr note_t minNote = 24; // C1
	static constexpr note_t maxNote = 95; // B6
	static constexpr juce::Range<note_t> validNoteRange{minNote, maxNote};
};
```


CrushedPixel:
This PR is onto the wrong branch - will reopen.

PR #460, opened 2019-01-04
Make Range constructor constexpr

CrushedPixel:
I needed the two-argument constructor for `juce::Range` to be `constexpr`, so that I can create `static const` class members inline:

```cpp
class Notes {
public:
	static constexpr note_t minNote = 24; // C1
	static constexpr note_t maxNote = 95; // B6
	static constexpr juce::Range<note_t> validNoteRange{minNote, maxNote};
};
```

### DIFF
--- a/modules/juce_core/maths/juce_Range.h
+++ b/modules/juce_core/maths/juce_Range.h
@@ -44,7 +44,7 @@ public:
     Range() = default;
 
     /** Constructs a range with given start and end values. */
-    Range (const ValueType startValue, const ValueType endValue) noexcept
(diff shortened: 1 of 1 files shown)


McMartin:
Did you make sure this compiles with VS2013?

CrushedPixel:
No, I'm developing on macOS and don't have a Windows Machine available ATM.
If someone could check this out, I would be very grateful.

ed95:
Yep, I've constexpr-ed a few Range methods in [84513ee](https://github.com/WeAreROLI/JUCE/commit/84513eef8f2b919822227a13eeb454aa7393a895).

CrushedPixel:
Thank you very much!

PR #468, opened 2019-01-18
Fix ppq based vst3 transport bug

getarndt:
projectTimeMusic == ppqPosition
Otherwise some sequencers does not work correctly (for example Rolands 808 (cloud))


seb-nektar:
Hi JUCE developers,
please consider merging this pull request. The bug leads to incorrect tempo syncing for some plugins and synths hosted in JUCE apps. We can confirm that the proposed PR fixes the problem.
Thanks! :)

ed95:
This was added to `develop` in [4126dc7](https://github.com/WeAreROLI/JUCE/commit/4126dc72b1d4a4b2bf8b83d69b9a213f3c897b03). It'll make its way onto `master` when we do a point release in the next couple of weeks.

PR #473, opened 2019-01-23
Cleanup: Remove assigned to but unused variables

McMartin:
A small cleanup that I found when looking at removing dependencies on `juce_graphics`, `juce_gui_basics` and `juce_gui_extra` from the `ResourceFile` class of Projucer.

@ed95, @julianstorer, @tpoole: merge at will, thanks!


ed95:
Added in [ee89aa8](https://github.com/WeAreROLI/JUCE/commit/ee89aa88986506769188e556e767c6f22fea8d54), thanks!

PR #483, opened 2019-02-08
Fix graphics performance: Don't need displayScale value changed when CoreGraphicsRenderer used.

COx2:
Hello, Geniuses!
I found a code that is a performance bottleneck when running on Retina Display with CoreGraphicsRenderer.

Actually, since NSHighResolutionCapable is set to true in Info.plist on Retina compatible applications, CoreGraphics seems to automatically scale internal back buffers by scale factor (factor is x2).
However, as I have seen, I recognize that there is a problem that JUCE side is doing affine transformation (scale multiplication).
I think that this implementation is the cause of the following problems that have been posted by JUCE Forum.
https://forum.juce.com/t/drawing-images-on-mbp-retina-is-ultra-slow/9548

On the other hand, when using SoftwareRenderer, the above scaling processing is necessary.

Based on the above, I propose an implementation that does not change displayScale value when using CoreGraphicsRenderer.

Pleasse check.

### DIFF
--- a/modules/juce_gui_basics/native/juce_mac_NSViewComponentPeer.mm
+++ b/modules/juce_gui_basics/native/juce_mac_NSViewComponentPeer.mm
@@ -849,9 +849,12 @@ public:
         float displayScale = 1.0f;
 
        #if defined (MAC_OS_X_VERSION_10_7) && (MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_X_VERSION_10_7)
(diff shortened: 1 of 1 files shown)


ed95:
Added in [90d6537](https://github.com/WeAreROLI/JUCE/commit/90d65375dfb22fd6c8bf950228a2e13f9804d8cd) - thanks!

COx2:
Thank you for your checking!

PR #484, opened 2019-02-08
Plug memory leak calling setStateInformation() on hosted VST3

getdunne:
See my JUCE Forum post https://forum.juce.com/t/memory-leak-calling-setstateinformation-on-hosted-vst3-plugin/31660, and test/demo project https://github.com/getdunne/juce-wrapper. I think the leak is caused by the `ComSmartPtr<Steinberg::MemoryStream>` object getting initialized with a reference count of 2 here, when it should be just 1.

### DIFF
--- a/modules/juce_audio_processors/format_types/juce_VST3PluginFormat.cpp
+++ b/modules/juce_audio_processors/format_types/juce_VST3PluginFormat.cpp
@@ -2289,7 +2289,7 @@ public:
 
         if (head != nullptr)
         {
-            ComSmartPtr<Steinberg::MemoryStream> componentStream (createMemoryStreamForState (*head, "IComponent"));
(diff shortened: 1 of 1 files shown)


tpoole:
For the second commit, what platform were you using when you spotted duplicates?

getdunne:
> For the second commit, what platform were you using when you spotted duplicates?

My mistake--that second commit wasn't meant to be part of the original memory leak fix. It's a separate issue. I committed the change to my own fork because I use GitHub to propagate among different development PCs and Macs. It didn't occur to me that it would come up for you as part of the open PR--sorry about that.

I don't consider this change definitive in any way, but the duplicates do occur, so probably ought to be investigated. 

To answer your question, I saw this on Windows 10 64-bit, when I opened a standard Audio Device Setup dialog and ticked, then un-ticked the box for input channels 1&2. For some reason the block size 128 was enumerated twice.

PR #488, opened 2019-02-18
Added FMA3 and FMA4 instruction detection, and improved Linux information support.

jrlanglois:


modules/juce_core/native/juce_linux_SystemStats.cpp
```diff
@@ -36,17 +36,16 @@ SystemStats::OperatingSystemType SystemStats::getOperatingSystemType()
 
 String SystemStats::getOperatingSystemName()
 {
+    auto result = readPosixConfigFileValue ("/etc/lsb-release", "DISTRIB_ID");
```
hjmallon: This will not work on all distros but then it will fall back to "Linux" so that seems OK. You should definitely be checking /etc/os-release first as that is supported on all recent systemd distros at least. Also this is probably a breaking change. https://www.freedesktop.org/software/systemd/man/os-release.html

modules/juce_core/native/juce_linux_SystemStats.cpp
```diff
@@ -36,17 +36,16 @@ SystemStats::OperatingSystemType SystemStats::getOperatingSystemType()
...
 
 bool SystemStats::isOperatingSystem64Bit()
 {
-   #if JUCE_64BIT
-    return true;
-   #else
-    //xxx not sure how to find this out?..
-    return false;
-   #endif
+    return (sizeof (void*) * CHAR_BIT) == 64;
```
hjmallon: When was the previous check not sufficient? Were you seeing issues on a certain device?

modules/juce_core/native/juce_mac_SystemStats.mm
```diff
@@ -71,6 +71,13 @@ static void doCPUID (uint32& a, uint32& b, uint32& c, uint32& d, uint32 type)
 //==============================================================================
 void CPUInformation::initialise() noexcept
 {
+   #if __FMA__
```
hjmallon: This define will be set if we are compiling with FMA/FMA4. This CPUInformation checks are about runtime. FMA is similar to SSE3 but bit 12 rather than 0. FMA4 is in a different page and will need a new  SystemStatsHelpers::doCPUID call.

Wikipedia has a good list but if you dont trust it double check with the Intel guide

modules/juce_core/native/juce_win32_SystemStats.cpp
```diff
@@ -144,6 +144,7 @@ void CPUInformation::initialise() noexcept
     hasSSE2  = (info[3] & (1 << 26)) != 0;
     hasSSE3  = (info[2] & (1 <<  0)) != 0;
     hasAVX   = (info[2] & (1 << 28)) != 0;
+    hasFMA3  = (info[2] & (1 << 12)) != 0;
```
hjmallon: You haven't checked for FMA4 here at all?

jrlanglois: Indeed. I haven't found a way to do it, even by following the official Microsoft example for [`__cpuid`](https://docs.microsoft.com/en-us/cpp/intrinsics/cpuid-cpuidex?view=vs-2017).  I'm all ears if you have any ideas!

modules/juce_core/native/juce_linux_SystemStats.cpp
jrlanglois: > You should definitely be checking /etc/os-release first as that is supported on all recent systemd distros at least.
Good ole Linux - immensely inconsistent and convoluted! Thanks, I'll add that check.

> Also this is probably a breaking change.
I don't see how this could be breaking in any way; all this does is (potentially) return a less generic string! If you can clarify how you see that as the case, that would be great. Even so, I don't see the downside here? The goal is to get provide better information.

modules/juce_core/native/juce_linux_SystemStats.cpp
jrlanglois: I'm trying to find a less `JUCE_*` macro-y way of doing it as it would be totally specific to JUCE itself. Like I mentioned on the forum post - open to any alternative. I think this case warrants a dynamic alternative, otherwise why bother with a platform specific solution in this function?

modules/juce_core/native/juce_linux_SystemStats.cpp
hjmallon: I agree the behavior is better. Just that people with code that says if `getOperatingSystemName() == "Linux"` will need to be informed that this is a breaking change.

modules/juce_core/native/juce_mac_SystemStats.mm
jrlanglois: > This define will be set if we are compiling with FMA/FMA4. This CPUInformation checks are about runtime.

Hm, good point and makes sense!

> FMA is similar to SSE3 but bit 12 rather than 0. FMA4 is in a different page and will need a new SystemStatsHelpers::doCPUID call.

Ahh, thanks! For the most part, I took this task on to learn more about how this all works... this is helping!

hjmallon: You should be able to call again with 0x80000001 then check bit 16 of c for fma4. It seems that extension is AMD only so almost no one will have that available (no AMD mac machines officially).

modules/juce_core/native/juce_win32_SystemStats.cpp
hjmallon: I would have a look in info[2] bit 16 of `callCPUID (info, 0x80000001)`

modules/juce_core/native/juce_linux_SystemStats.cpp
jrlanglois: I see what you mean. It's really up to Roli, but people should be reminded to be using [`juce::SystemStats::getOperatingSystemType()`](https://github.com/WeAreROLI/JUCE/blob/master/modules/juce_core/system/juce_SystemStats.h#L74-L79) instead of string based checks.

modules/juce_core/native/juce_mac_SystemStats.mm
jrlanglois: Is there a risk of borking anything when running on iOS when performing such a call? I don't have a way to test that platform to be sure the system isn't throwing red-flags of any kind. I just know that such a highly sandboxed system tends to be sensitive.

hjmallon: The block that checks them just below is inside a `#if JUCE_INTEL && ! JUCE_NO_INLINE_ASM` so it will not be done on iOS

jrlanglois: Nice, good to know

modules/juce_core/native/juce_linux_SystemStats.cpp
hjmallon: That sizeof() is checking at compile time. Try building a test program with `gcc -m32 foo.c` on 64bit Linux to check. In that case (compiling 32bit and running on 64bit OS) it will return 32bit. 

You may want to try `uname()` and looking at the machine string. But then you will need to make a list of possible JUCE supported `uname -m` strings and whether they are 64bit. This becomes difficult if it needs updating in future as new ARM arch are introduced for instance.

jrlanglois: Good point. I'll dig around to see if there's a list of example return types and get rolling on fixing this up...

jrlanglois: @hjmallon If I've done my homework correctly, the new commit I pushed should satisfy the majority of cases. Let me know what you think.

modules/juce_core/native/juce_win32_SystemStats.cpp
jrlanglois: Changed, and pushed.

modules/juce_core/native/juce_mac_SystemStats.mm
jrlanglois: @hjmallon Where did you find that detail, btw?

modules/juce_core/native/juce_linux_SystemStats.cpp
hjmallon: I would be tempted to leave the #ifdef JUCE_64BIT, return true section in as that will always (until JUCE supports 128bit OSs) be true and that prevents any regressions here.

jrlanglois: Since the goal is to dynamically check this detail, I don't see what's wrong with this route that I've taken at this point. Besides, this would (at best) land in `develop` which is _the_ place to catch regressions. And I highly doubt JUCE will support a 128bit OS - they're barely useful enough to matter as it is...

> Also check the return value

Mm... why bother? Worst case here is that `machine` is null or empty, which will fail all conditions.

jrlanglois: I see an improvement already where the case should be ignored.

PR #489, opened 2019-02-22
Enable HeavyweightLeakedObjectDetector to be used for DynamicObject

jamiebullock:
A very useful application of HeavyweightLeakedObjectDetector is tracing the sources of leaks for DynamicObjects. This simple patch removes the dependency of `HeavyweightLeakedObjectDetector` on `HashMap` so it can be used for this purpose.


ed95:
Added in [09daeb3](https://github.com/WeAreROLI/JUCE/commit/09daeb3a6d7df0e163345d66e2ad27e64b796772), thanks!

kunitoki:
wouldn't have been faster to use an unordered_map ?

PR #495, opened 2019-03-02
Support browsing for multiple directories

McMartin:
Feature Request: https://forum.juce.com/t/filechooser-browseformultipledirectories/32027
Original commit by @asimilon: https://github.com/asimilon/JUCE/commit/a341225cf80abe0912f986cae8e11af6f508ae53


McMartin:
Closing since according to @ed95 
> This only works correctly on macOS, it’s not currently possible with the native Windows or Linux file browsers which is probably why the method is missing from the `FileChooser` API.

https://forum.juce.com/t/filechooser-browseformultipledirectories/32027/5?u=mcmartin

PR #497, opened 2019-03-02
Fix AudioDeviceManager::setAudioDeviceSetup to check for different number of channels needed

RobJellinghaus:
My app needs lowest latency on Windows, and I have been experimenting with the various Windows audio devices.  Unsurprisingly ASIO wins by a landslide.  So I want my app to start up immediately with an ASIO device created.

I found that when calling initialise() followed by setCurrentAudioDeviceType(), there was a 1.5 second pause due to the Thread::sleep call "to help avoid things like DirectSound/ASIO clashes".  But in my case I know I want ASIO when my app starts up, so I don't want this pause.

Modifying the JUCEAudioInputProcessor sample to use an AudioProcessorGraph, I wound up with the initialization code in [the prepareToPlay() method](https://github.com/RobJellinghaus/JuceProcessingAudioInputGraphTutorial/blob/startup_with_asio/Source/ProcessingAudioInputTutorial.h):

```
        if (desiredTypeName.length() > 0)
        {
            deviceManager.setCurrentAudioDeviceType(desiredTypeName, /*treatAsChosenDevice*/ false);
        }
        else
        {
            throw std::exception("Could not set audio device type to desired type name");
        }

        String result = deviceManager.initialiseWithDefaultDevices(2, 2);
        if (result.length() > 0)
        {
            throw std::exception(result.getCharPointer());
        }
```

Unfortunately this did not work; the number of input channels created wound up being zero, rather than 2 as requested by the initialiseWithDefaultDevices call.

It turns out that what was going wrong was:

- AudioDeviceManager::AudioDeviceManager() has initializers for numInputChansNeeded (0), and numOutputChansNeeded (2).
- My code calls setCurrentAudioDeviceType("ASIO", false).
  - This winds up calling AudioDeviceManager::setAudioDeviceSetup.
  - That setup call has an "if" clause at line 476 which checks if device names changed or there is no current device.
    - Since there is no current device, the if branch is taken.
    - This initializes AudioDeviceManager::inputChannels to (bitset value) 0 and AudioDeviceManager::outputChannels to (bitset value) 3, which is correct given the existing constructed values of num{In,Out}putChansNeeded.
- My code then calls initialiseWithDefaultDevices(2, 2).
  - This again winds up calling AudioDeviceManager::setAudioDeviceSetup.
  - However, this time, the "if" clause at line 476 is *not* taken.
    - This is because the ASIO device has already been set up and the input and output device names have not changed.

The result: even though the device manager's numInputChansNeeded is finally set to 2, the device manager's inputChannels bitset is still at 0, preventing input.

This pull request is the minimal fix I thought appropriate for this specific issue:  if the number of channels needed differs from the number offered by the device, recreate the device (as though the device name had changed).

With this patch, my JUCE audio input processor graph sample app runs properly with ASIO (and beautifully low latency):

https://github.com/RobJellinghaus/JuceProcessingAudioInputGraphTutorial/tree/startup_with_asio/Source


RobJellinghaus:
@ed95 fixed this in a slightly different and overall better way, [see this JUCE forum thread for details.](https://forum.juce.com/t/wasapi-exclusive-or-asio-devices-have-no-input-channels-why/31482/4)  So, closing this PR as the problem is now fixed.  Thanks Ed!

PR #510, opened 2019-03-25
Implement BluetoothMidiDevicePairingDialogue for Mac OS X

dingari:
This PR utilizes the [CABTLEMIDIWindowController](https://developer.apple.com/documentation/coreaudiokit/cabtlemidiwindowcontroller?language=objc) class from CoreAudioKit to implement [BluetoothMidiDevicePairingDialogue](https://docs.juce.com/master/classBluetoothMidiDevicePairingDialogue.html) for Mac OS X systems (SDK 10.11+).

This can be tested by building the Audio Plugin Host from the extras folder. The "Audio Settings" window will contain a button that opens this dialogue.

There are  a few things to note:
* My Objective C experience is limited, and there may be room for improvements there. Just let me know if you spot anything obvious.
* Since the CABTLEMIDIWindowController class simply spawns a new window, this isn't really a "modal" component, but I made it follow the same API as the iOS version, taking a modal callback object that is called when the window is closed.
* I'm a little concerned about the `BluetoothMidiDevicePairingDialogue::open` function. It simply calls `new BluetoothMidiSelectorOverlay (cb.release(), boundsToUse);` and exits. Where is this instance deleted? Is this done automatically when the "modal" state is exited? I had to go around this by using `MessageManager::callAsync` to trigger deletion of the object to avoid memory leaks.

### DIFF
--- a/modules/juce_audio_utils/native/juce_mac_BluetoothMidiDevicePairingDialogue.mm
+++ b/modules/juce_audio_utils/native/juce_mac_BluetoothMidiDevicePairingDialogue.mm
@@ -24,22 +24,122 @@
   ==============================================================================
 */
 
+#include <CoreAudioKit/CABTLEMIDIWindowController.h>
+
+// This is a very basic wrapper class that can open a Bluetooth MIDI pairing dialogue
+// on Mac OS X and notify it's owner when the window is being closed.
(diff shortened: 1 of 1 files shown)


tpoole:
Thanks for this.

Launching the macOS pairing dialogue is possible on the `develop` branch.

PR #511, opened 2019-03-30
Added warning about overwriting files to zenity 

B-Raffai:
Zenity is used as the file chooser on Linux when KDE is not running. Previously, when an existing file was selected while saving a file the existing file would be overwritten with no warning. Now, there will be an appropriate warning message.


ed95:
Added in [fec3b7d](https://github.com/WeAreROLI/JUCE/commit/fec3b7d59ed5693b27f476880becc32243d4f8b2).

Issue #514, opened 2019-04-02
Projucer: Android extra resource files are handled correctly


MatthieuOrosys:
The fields 
  * Extra Android XML Value Resources
  * Extra Android Drawable Resources
  * Extra Android Raw Resources

are not usable.
Whatever I put in this filed (relative path, absolute path), it does not work.

After a bit of code review, it seems that there is an issue in

    AndroidProjectExporter::copyExtraResourceFiles (const String& resources, const String& dstRelativePath)

The line

    auto file = getProject().getFile().getChildFile (path);

should be

    auto file = getProject().getFile().getParentDirectory().getChildFile (path);

PR #515, opened 2019-04-02
Projucer exporter for Android: Fix path for Extra resources

MatthieuOrosys:
Path was constructed from the project file itself instead of its parent directory.


MatthieuOrosys:
See #514

PR #521, opened 2019-04-25
Windows: Fixed double delete crash in WebBrowserComponent::goToURL() when invoked with POST data

matok:
Calling both `VariantClear (&postDataVar)` and `SafeArrayDestroy (sa)` at the end of win version of `WebBrowserComponent::goToURL()` causes crash after double deleting.

Can be easily reproduced by calling `WebBrowserComponent::goToURL()` with some POST data.


ed95:
Added in [989d860](https://github.com/WeAreROLI/JUCE/commit/989d86042808de0cddee851efad34239c9e24ad2) - thanks!

PR #523, opened 2019-05-02
Add minimum "GUI-less" Linux VST3 support. (host only)

atsushieno:
This partly addresses https://github.com/WeAreROLI/JUCE/issues/307 (which is in turn referenced at https://forum.juce.com/t/vst3-support-on-linux/31872/2). The changes aim to bring VST3 to JUCE apps on Linux. This does NOT bring GUI support, but the audio part seems to be working. This seems to help some apps that don't depend on juce_gui_basic. Maybe only partially, but AudioPluginHost successfully listed and loaded some VST3 plugins (mda-vst for example).

VST3 plugins listed and loaded on AudioPluginHost

### DIFF
--- a/extras/Projucer/Source/ProjectSaving/jucer_ProjectExport_Make.h
+++ b/extras/Projucer/Source/ProjectSaving/jucer_ProjectExport_Make.h
@@ -179,6 +179,7 @@ public:
                 case VSTPlugIn:
                 case UnityPlugIn:
                 case DynamicLibrary:        return ".so";
+                case VST3PlugIn:            return ".vst3"
                 case SharedCodeTarget:
                 case StaticLibrary:         return ".a";
                 default:                    break;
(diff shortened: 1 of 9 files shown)


atsushieno:
It was host only, but I could apply the changes to tracktion_engine and successfully play my sample tracktionedit song data using its PlaybackDemo. https://gist.github.com/atsushieno/8325f7bd3a35131d8250f3087c55b8f2

tpoole:
Thank you for contributing your changes, but we've currently got more comprehensive VST3 support in the works behind the scenes. I'm afraid I can't give a firm date for when it will be released publicly though.

atsushieno:
I'm totally fine with that and that's my expectation. The PR is here mostly for those who are looking for solution to current "complete lack of VST3 support" situation and not necessarily for merging. I for one definitely want GUI-less VST3 support to make use of tracktion_engine.

I also figured that it is not necessarily JUCE issue but rather blocked at steinbergmedia/vst3sdk e.g. https://github.com/steinbergmedia/vst3sdk/issues/34 , as well as various "not implemented" bits there (e.g. in fstring). I'd expect ROLI team can handle the entire situation in much better way.

atsushieno:
This does not have to be here anymore right? I'm closing this and looking forward to seeing juce6!

PR #528, opened 2019-05-13
add option in JUCE to allow forbidden characters in OSC addresses.

benkuper:
I've been struggling with a lot of people and software designers who decided to prioritize "user-friendly" osc addresses (e.g. with spaces : /my super/address) against OSC specification compliance. Those are famous software in digital art, like MadMapper or Millumin.
The way OSC is handled in JUCE right now doesn't allow programmers to decide whether they want to allow those addresses, and it can be a blocker if you really need to be able to communicate with those not standard softwares.

The proposal is to add a module option (disabled by default), that bypasses the special chars check when enabled

### DIFF
--- a/modules/juce_osc/juce_osc.h
+++ b/modules/juce_osc/juce_osc.h
@@ -54,6 +54,14 @@
 #include <juce_core/juce_core.h>
 #include <juce_events/juce_events.h>
 
+/** Config: JUCE_ALLOW_SPECIAL_CHARS_IN_ADDRESS
+	Enables the use of characters in adress that are not allowed by the OSC specifications (like spaces), but that are used
+	by some applications anyway (e.g. /my spaced/address)
+*/
+#ifndef JUCE_ALLOW_SPECIAL_CHARS_IN_ADDRESS
+#define JUCE_ALLOW_SPECIAL_CHARS_IN_ADDRESS 0
+#endif
+
 //==============================================================================
 #include "osc/juce_OSCTypes.h"
 #include "osc/juce_OSCTimeTag.h"
--- a/modules/juce_osc/osc/juce_OSCAddress.cpp
+++ b/modules/juce_osc/osc/juce_OSCAddress.cpp
@@ -274,10 +274,17 @@ namespace
             return c >= ' ' && c <= '~';
         }
 
+#pragma warning(push)
+#pragma warning(disable: 4100)
         static bool isDisallowedChar (juce_wchar c) noexcept
         {
+#if JUCE_ALLOW_SPECIAL_CHARS_IN_ADDRESS
+			return false;
+#else
             return CharPointer_ASCII (Traits::getDisallowedChars()).indexOf (c, false) >= 0;
+#endif
         }
+#pragma warning(pop)
 
         static bool containsOnlyAllowedPrintableASCIIChars (const String& string) noexcept
         {


FigBug:
This would be very useful. TouchOSC illegal characters in it's default maps.

designerfuzzi:
OSC protocol depends on data alignment. If the space and rest of string fits in alignment then it would still be ok and part of the string. Otherwise how do you send a simple string `"hello world"`? 
Your suggestion that this would be useful would force to be able to distinguish between address decoding where spaces would be not allowed and string decoding of messages where it should be allowed for each packet.

It is not user convenience, it is bug resistance. A space character is easily send in the osc address no matter if that was intentional or not. Also `00` != `\0` != ` ` are different data, which is why the alignment rule in OSC protocol exists that applies also to address strings.

sending special characters leads to sending unicodes, which have different alignment and which is not part of the protocol.

This is the reason why most osc code encodes and decodes from receiving and to sending, 
a) to fullfil aligmment 
b) to escape not allowed characters 
c) to block out malicious attempts of send data, which is specially needed since a lot OSC implementations also are publishing their port DNS records via Bonjour (actually the older zeroconf).

benkuper:
Thank you for the very thorough explanation.
How do you suggest I should do that while still using the juce_osc module ? If the proposed solution means "deal with your own OSC implementation", then I don't find it very constructive.. 
If there is a "JUCE friendly" way to still use the juce_osc module but being able to receive and parse those adresses, then I would love to discuss how and propose a PR going this direction.

designerfuzzi:
You just need to tinker a proper encoding method/function that allows you to put the data you need into the 4byte alignment scheme of OSC readable again on receiver side in the decoder function. Nothing superstitious. 
You might need to specify what your forbidden characters are because there are a looot different ways to encode and decode such. In other words, when the idea is to step out of the standard given by stanford osc1.0 then what makes you believe data can't be send with that standard? 
A space character in 4byte alignment is a single "00" field. No matter where it appears in the full message, the message is broken down into 4 byte chunks and parsed. The difficulty is only that a space at the end of a 4byte chunk and same time end of a message part (like the address) looks like you could ignore the 00 and split the string right before. So your decoder needs to handle chunks accordingly when some character follows a 00 byte up to its alignment end. Which actually is also the standard given as far i understood the papers. 
In case of unicode characters which need at least 4 byte to be stored this looks different because a following 4 byte alignment chunk might be a unicode as well.. which forces you to be able to tell your encoder as well as the decoder what encoding is used. 
Hint: Juce has its own string model encoding and decoding, while OSC standard is given as ascii sequence. 
Juce strings are not ascii strings. Means you still end up rewriting decoder and encoder for OSC using juce strings.

benkuper:
I totally agree that steppign out of the standard is NOT a good idea and the real fix should be done on the other side.
But I don't have access to the [many] software that don't respect it, and my goal developping the software I build is to be as compliant as possible with all of them, still trying to "educate" people on respecting the standard.
I thought it would be a nice addition to JUCE to allow that in a simple way that proves working for all the cases I've encountered, but I guess this would need way more work to have something that you would be happy with, unfortunately I don't have that time :)
FYI, this one of the main projects I have had the need for that : https://benjamin.kuperberg.fr/chataigne 

Cheers and thank you for taking the time to explain

PR #530, opened 2019-05-17
Optimization of VST3 shell plug-ins instantiation

fredmerging:
Instantiating a VST3 shell plug-in can be overkill. For example, if using Waves complete suite of plug-ins (more than 400 plug-ins), the instantiation of the first enumerated plug-in (e.g. Waves API-2500) is almost instant, while the instantiation of the bottom listed plug-ins (e.g. Waves Q10) can take more than a minute.
This code update avoids creating an instance of each component handled by a vst3 factory when opening a VST3 module, and instead just checks if the name of the provided description can be found in the enumerated classes of a vst3 factory. It allows a huge speed-up of plug-in instantiation for big VST3 shells such as Waves.

### DIFF
--- a/modules/juce_audio_processors/format_types/juce_VST3PluginFormat.cpp
+++ b/modules/juce_audio_processors/format_types/juce_VST3PluginFormat.cpp
@@ -1104,9 +1104,9 @@ private:
     Result findName(IPluginFactory* factory, const PluginDescription& description)
     {
         if (factory == nullptr)
-            return Result::fail(L"Invalid Parameter");
+            return Result::fail("Invalid Parameter");
 
-        Result result(Result::fail(L"not found"));
+        Result result(Result::fail("not found"));
         auto numClasses = factory->countClasses();
 
         for (Steinberg::int32 i = 0; i < numClasses; ++i)
--- a/modules/juce_audio_processors/format_types/juce_VST3PluginFormat.cpp
+++ b/modules/juce_audio_processors/format_types/juce_VST3PluginFormat.cpp
@@ -1089,11 +1089,9 @@ private:
         if (pluginFactory != nullptr)
         {
             ComSmartPtr<VST3HostContext> host (new VST3HostContext());
-            MatchingDescriptionFinder finder (host, pluginFactory, description);
 
-            auto result = finder.findDescriptionsAndPerform (f);
-
-            if (result.getErrorMessage() == MatchingDescriptionFinder::getSuccessString())
+            auto result = findName(pluginFactory, description);
+            if (result.wasOk())
             {
                 name = description.name;
                 return true;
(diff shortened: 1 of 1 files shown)


modules/juce_audio_processors/format_types/juce_VST3PluginFormat.cpp
```diff
@@ -1103,6 +1101,30 @@ struct VST3ModuleHandle  : public ReferenceCountedObject
         return false;
     }
 
+    Result findName(IPluginFactory* factory, const PluginDescription& description)
+    {
+        if (factory == nullptr)
+            return Result::fail(L"Invalid Parameter");
```
jrlanglois: > `L""`

Why are you adding `L`? It's not necessary at all.

modules/juce_audio_processors/format_types/juce_VST3PluginFormat.cpp
```diff
@@ -1103,6 +1101,30 @@ struct VST3ModuleHandle  : public ReferenceCountedObject
...
+
+        Result result(Result::fail(L"not found"));
+        auto numClasses = factory->countClasses();
+
+        for (Steinberg::int32 i = 0; i < numClasses; ++i)
+        {
+            PClassInfo info;
+            factory->getClassInfo(i, &info);
+
+            if (std::strcmp(info.category, kVstAudioEffectClass) != 0)
```
jrlanglois: I vaguely recall that many audio plugins don't support this method... I suggest testing with a variety of open-source plugs to be sure.

jrlanglois: Have a look at `DescriptionFactory` to get a better sense of it.

modules/juce_audio_processors/format_types/juce_VST3PluginFormat.cpp
fredmerging: Sorry, too used to code for Windows Unicode. will remove it

modules/juce_audio_processors/format_types/juce_VST3PluginFormat.cpp
fredmerging: I picked this piece of code from DescriptionFactory. Basically this findName() method I added is the same as DescriptionFactory::findDescriptionsAndPerform() but I removed anything that is not needed. So the criteria are the same as before except we don't instantiate and destroy the component.
What I want to check in this loop is: 
  - the name of the component (does it match the the provided one nested in the passed description)
  - is it an audio processor (we ignore the user interface component)

modules/juce_audio_processors/format_types/juce_VST3PluginFormat.cpp
fredmerging: done

tpoole:
Thank you.

Added in https://github.com/WeAreROLI/JUCE/commit/9c50bbde2ba415697013902e213265b3b35cc975.

PR #531, opened 2019-05-19
Added ControllableProcessorBase to recycle AudioProcessorParameters

ffAudio:
In order to control other Processors in a processing pipeline, a separate interface is needed, so the AudioProcessorParameters, the AudioProcessorValueTreeState and the GenericAudioProcessorEditor can be used. A use case is e.g. a VideoProcessor in a NLE.

This PR adds this base `ControllableProcessorBase` and moves all addParameter etc. functionality into this bae.
The AudioProcessorParameter was changed to use the interface of ControllableProcessorBase instead of AudioProcessor.
The GenericAudioProcessorEditor uses the LegacyAudioParameter and the AudioProcessorListener, so it remains unchanged (but intact of course). A new GenericProcessorEditor will be added later, that will only use the new interface of AudioProcessorParameter and AudioParameterGroup as well.

If wished, the names AudioProcessorParameter and AudioProcessorValueTreeState could be changed for consistency to ProcessorParameter and ProcessorValueTreeState and an alias provided with the old names.

Disclaimer: all copyright of this PR is hereby waived to JUCE and its legal owners.

### DIFF
--- a/modules/juce_audio_processors/processors/juce_AudioProcessorEditor.cpp
+++ b/modules/juce_audio_processors/processors/juce_AudioProcessorEditor.cpp
@@ -27,12 +27,12 @@
 namespace juce
 {
 
-AudioProcessorEditor::AudioProcessorEditor (AudioProcessor& p) noexcept  : processor (p)
+AudioProcessorEditor::AudioProcessorEditor (ControllableProcessorBase& p) noexcept  : processor (p)
 {
     initialise();
 }
 
-AudioProcessorEditor::AudioProcessorEditor (AudioProcessor* p) noexcept  : processor (*p)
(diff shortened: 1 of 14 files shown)


modules/juce_audio_processors/processors/juce_AudioProcessorEditor.cpp
```diff
@@ -45,7 +45,7 @@ AudioProcessorEditor::~AudioProcessorEditor()
 
     // if this fails, then the wrapper hasn't called editorBeingDeleted() on the
     // filter for some reason..
-    jassert (processor.getActiveEditor() != this);
+//    jassert (processor.getActiveEditor() != this);
```
ffAudio: Oops, I'll fix that asap

ffAudio:
WontBeAdded

PR #537, opened 2019-06-03
Fix a potential X11 BadWindow error

jpcima:
Hi, this fixes a somewhat rare X window error which may crash plugin and host.

The Xlib function [XGetInputFocus](https://tronche.com/gui/x/xlib/input/XGetInputFocus.html) may return the special result `PointerRoot` which is neither 0 not a valid Window ID.
When this is passed to `XQueryTree` by the function call `isParentWindowOf` following, it will raise the error, so it's necessary to take special care this particular case.


ed95:
Added in [022e2e2](https://github.com/WeAreROLI/JUCE/commit/022e2e2bea44250ab05f6b24cc4c1b06a314b539) - thanks for reporting!

PR #540, opened 2019-06-21
Add visibility status in ResizableWindow state get/restore methods

sigmate:
A small PR to easily also save/restore a window's visibility (e.g. supposably when using `ApplicationProperties` during a GUI application initialisation or shutdown).

The `forceVisible` param is handy to ensure a "main" window would always be visible, thus avoiding the potential confusion for the user to not see any opened window upon starting the application).

Example:
```
mainWindow->restoreWindowStateFromString (applicationProperties.getUserSettings()->getValue ("mainWindowState"), true);
        
auxWindow->restoreWindowStateFromString  (applicationProperties.getUserSettings()->getValue ("auxWindowState"));
```

**Note**: in order not to break anything with previously saved window states, the `restoreWindowStateFromString` method will look for a "h" (for hidden) token in the string.

### DIFF
--- a/modules/juce_gui_basics/windows/juce_ResizableWindow.cpp
+++ b/modules/juce_gui_basics/windows/juce_ResizableWindow.cpp
@@ -528,18 +528,20 @@ void ResizableWindow::parentSizeChanged()
 String ResizableWindow::getWindowStateAsString()
 {
     updateLastPosIfShowing();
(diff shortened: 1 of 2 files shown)


sigmate:
Sorry, just spotted a little issue. Will fix it by the end of the day.

sigmate:
Ok, done.

PR #547, opened 2019-07-04
add setBlockingState in read function, like DatagramSocket read function

benkuper:
When doing a read() on a StreamingSocket with shouldBlock set to false, it will still be blocking : the debug becomes untraceable at juce_Socket.cpp line 213 (goes to native code I guess) :

``` bytesThisTime = ::recv (handle, buffer, numToRead, 0) ```

 but the code won't process until at least one byte is received.

A post on the forum suggest this modification and it works, maybe not the best but as it is used the same in DatagramSocket, I guess it's not that bad.

i'm on Win10 x64


ed95:
This has been added in [1baa825](https://github.com/WeAreROLI/JUCE/commit/1baa825f2ceaa536eb1b858f1e9ae41b4a1111e3).

PR #548, opened 2019-07-05
Linux: fix scaling factor computation

piem:
Correcting the constant value from 150 to 96, as mentioned in the comment, was enough for me to get the resolution fixed on a HiDPI screen.


ed95:
Added in [e206bbe](https://github.com/WeAreROLI/JUCE/commit/e206bbecc4faace98b0eac3fc4789287aa0133d6).

piem:
hrm, thanks @ed95.

PR #551, opened 2019-07-11
fixes delivery of UTF-8 encoded strings via OSC

beschulz:
When sending strings via OSC that contain non-ASCII characters, "OSC input stream format error: missing padding zeros" was thrown in the OSCReceiver. This patch fixes this problem.


tpoole:
Thanks, added in https://github.com/WeAreROLI/JUCE/commit/14a24d614fbd4f6f146832ab1d50f2525a550bfe.

PR #552, opened 2019-07-11
Added float CallOutBox::LookAndFeelMethods::getCallOutBoxCornerSize(c…

beschulz:
…onst CallOutBox&) so that the corner radius of the CallOutBox bubble can be customised and an implementation in LookAndFeel_V2 that preserves the current behaviour.


tpoole:
Thanks. Added in https://github.com/WeAreROLI/JUCE/commit/3c9fe89636553484f93b9729e6ca8974a16a01b2.

PR #568, opened 2019-08-08
Feature/fix clamping of image convolution kernel

beschulz:
@julianstorer 
1.) a wrong pixel stride was used for single channel images
2.) The RGB and SingleChannel versions of the filter did not properly clip the output resulting in display artefacts.


beschulz:
@julianstorer : here's a depiction of the overflow. Notice the black specs in the corners of the wiggly line.

image

PR #571, opened 2019-08-08
Fixes inefficient repainting on iOS.

beschulz:
CGContextGetClipBoundingBox() does NOT return the area requested by "[view setNeedsDisplayInRect ...]". This commit implements handling of our own dirty rect list to rectify this problem.

Also see https://forum.juce.com/t/solved-repaint-ignores-opacity-and-repaints-parent-component/33711/16

### DIFF
--- a/modules/juce_gui_basics/native/juce_ios_UIViewComponentPeer.mm
+++ b/modules/juce_gui_basics/native/juce_ios_UIViewComponentPeer.mm
@@ -326,6 +326,8 @@ private:
                 peer->repaint (rect);
         }
     };
+    
+    RectangleList<int> dirtyRects;
 };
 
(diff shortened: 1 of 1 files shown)


beschulz:
Note: The whole approach is based on the (most likely false) assumption, that the CALayer of the UIView is ALWAYS caching the contents of the last draw (and is always the same). To leave undefined behaviour land one needs to draw into an image based context (via CGBitmapContextCreateImage()) and display this instead.

julianstorer:
Yeah, I played around with this whole issue when it first appeared a few years ago but couldn't find any way to actually work around it in a watertight way. I suspect that without forcing the entire area to be cached then it's simply impossible without access to the OS's underlying clip region.

beschulz:
see https://github.com/WeAreROLI/JUCE/pull/572

PR #574, opened 2019-08-13
Android: added missing method handling for ActivityLifecycleCallbacks…

essej:
Needed to prevent launch crash when running on Android Q devices


ed95:
Added in [be8a568](https://github.com/WeAreROLI/JUCE/commit/be8a568747854b3263b22a6996938009d09dbcff) - thanks!

PR #578, opened 2019-08-27
Update juce_PerformanceCounter.cpp

leighmarble:
Only printStatistics on destruction if there's something to print.

Otherwise, there's always an extra log entry added, that looks like:

Performance count for "my test" over 0 run(s)
Average = 0 microsecs, minimum = 0 microsecs, maximum = 0 microsecs, total = 0 microsecs

### DIFF
--- a/modules/juce_core/time/juce_PerformanceCounter.cpp
+++ b/modules/juce_core/time/juce_PerformanceCounter.cpp
@@ -43,7 +43,7 @@ PerformanceCounter::PerformanceCounter (const String& name, int runsPerPrintout,
 
 PerformanceCounter::~PerformanceCounter()
 {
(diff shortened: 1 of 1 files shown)


leighmarble:
Well, let me clarify my comment on this commit: it doesn't _always_ add an extra log entry. But if you are looping over something N times, and have the PerformanceCounter's runsPerPrintout = N, then the PerformanceCounter exits with 0 runs, and adds that extra (empty) log entry.

ed95:
That sounds sensible, I've added this in [e973451](https://github.com/WeAreROLI/JUCE/commit/e973451a73a8f44ded8103a67ec4ea1d2d09d15e).

PR #589, opened 2019-09-13
fix: Fix detecting au-plugins without 'thng' resource

abrykt:
Juce plugins includes .rsrc file which includes 'thng' identifier.
This fix enables detecting plugins that does not have .rsrc file.

See https://forum.juce.com/t/audio-units-without-thng-resource-how-to-load-from-file/28031


ed95:
Added in [4635e91](https://github.com/WeAreROLI/JUCE/commit/4635e91603cca2678463dca3f4b1b05f634b371c) - thanks for reporting!

abrykt:
Great! 👍

PR #591, opened 2019-09-16
Fix -Wreturn-std-move-in-c++11 warnings

danra:
Full Xcode 11 warning (for `splice`):

> warning: prior to the resolution of a defect report against ISO C++11, local variable 'a' would have been copied despite being returned by name, due to its not matching the function return type ('juce::var' vs 'Array\<juce::var\>') [-Wreturn-std-move-in-c++11]


ed95:
Added in [a24341d](https://github.com/WeAreROLI/JUCE/commit/a24341d2a752c066e4e9bb5f8f2b20cc9004bb55) - thanks for reporting!

PR #593, opened 2019-09-23
Fixed the path for precompiled libraries when using Visual Studio

McMartin:
I found this small discrepancy when working on https://github.com/McMartin/FRUT/pull/523.

@ed95 @tpoole FYI


tpoole:
Thank you. I'll get that added.

PR #598, opened 2019-10-02
Fixed MSVC warnings when using generated code from Projucer's component editor

codecat:
This fixes this warning from being spammed when building files generated with this code.

```
warning C4244:  'argument': conversion from 'int' to 'ValueType', possible loss of data
warning C4244:         with
warning C4244:         [
warning C4244:             ValueType=float
warning C4244:         ]
```


ed95:
Added in [d332fcc](https://github.com/WeAreROLI/JUCE/commit/d332fcc3c3d7b7df1d11d33e4228e1d6879651ca), thanks!

PR #604, opened 2019-10-10
Fix compiling juce_core/network/juce_Socket.cpp with MinGW

McMartin:
Compiling `juce_core/network/juce_Socket.cpp` stopped working with MinGW
in b857f965ce85aa682f3a3475341026b1b25a1fb2 because `WSAPOLLFD` requires
Windows Vista (or Windows Server 2008) (see https://docs.microsoft.com/en-us/windows/win32/api/winsock2/ns-winsock2-wsapollfd#requirements).

Setting `_WIN32_WINNT` to `0x501` means targeting Windows XP (and Windows
Server 2003), while setting it to `0x600` means targeting Windows Vista
(and Windows Server 2008) (see https://docs.microsoft.com/en-us/windows/win32/winprog/using-the-windows-headers) .

@ed95 please merge, thanks!

### DIFF
--- a/modules/juce_core/native/juce_BasicNativeHeaders.h
+++ b/modules/juce_core/native/juce_BasicNativeHeaders.h
@@ -126,7 +126,7 @@
  #define STRICT 1
  #define WIN32_LEAN_AND_MEAN 1
  #if JUCE_MINGW
-  #define _WIN32_WINNT 0x0501
+  #define _WIN32_WINNT 0x0600
  #else
   #define _WIN32_WINNT 0x0602
  #endif


ed95:
Thanks @McMartin. I've added back in the `select()` calls for MinGW instead in [https://github.com/WeAreROLI/JUCE/commit/170ec47a5e3d0ee40763b02ae9a1d4a659331a26](170ec47) as I'm wary of bumping the WINNT macro in case it breaks other things.

McMartin:
@ed95 https://github.com/WeAreROLI/JUCE/commit/170ec47a5e3d0ee40763b02ae9a1d4a659331a26 doesn't compile with MinGW:
```
D:/dev/JUCE/modules/juce_core/network/juce_Socket.cpp: In function 'int juce::SocketHelpers::waitForReadiness(std::atomic<int>&, juce::CriticalSection&, bool, int)':
D:/dev/JUCE/modules/juce_core/network/juce_Socket.cpp:344:13: error: 'result' was not declared in this scope; did you mean 'Result'?
  344 |         if (result < 0)
      |             ^~~~~~
      |             Result
```
I'll open another PR with a fix.

PR #610, opened 2019-10-16
Avoid AudioUnitSetProperty for kAudioUnitProperty_AudioChannelLayout on AU plugins that don't support it

seb-nektar:
This fixes a crash e.g. when loading the FabFilter One plugin in the the JUCE AudioPluginHost.

**Patch description**

Although it was previously checked in line 767 that kAudioUnitProperty_AudioChannelLayout is not supported, it might happen that it is set in line 790. To avoid this, move the AudioUnitSetProperty() call into the previous

`if (supportsLayouts) {...}`

section.


ed95:
Added in [2b542dc](https://github.com/WeAreROLI/JUCE/commit/2b542dc701e45cc9dcceb295cdffb8bb4b5fe22a) - thanks for reporting!

PR #611, opened 2019-10-16
VST3 plugin should not report any MIDI controller assignments when JUCE_VST3_EMULATE_MIDI_CC_WITH_PARAMETERS=0

seb-nektar:
When JUCE_VST3_EMULATE_MIDI_CC_WITH_PARAMETERS=0 the current behaviour is undefined. No Midi mappings should be reported when MIDI CC emulation is disabled.

**Proposed change:** Add #ifdef guard to check if MIDI CC is mapped to parameters, otherwise return kResultFalse to indicate that there are no mappings.


ed95:
Added in [a47dc92](https://github.com/WeAreROLI/JUCE/commit/a47dc92f51097368347db5527aa20ecf8277b941) - thanks for reporting!

PR #613, opened 2019-10-16
VST2 synth plugins always want MIDI input

seb-nektar:
Some VST2 synth plugins report that they do not want MIDI input. But any synth needs MIDI input so one can implicitly assume that every VST2 synth wants MIDI.

If `wantsMidiMessages` is false, no MIDI data is forwarded in the VST2 wrapper code and the plugin does not receive any input and cannot be used. We encountered this problem with the KORG Gadget VST2 plugins on OSX but probably other plugins show the same behaviour. See this [JUCE forum thread](https://forum.juce.com/t/problems-with-some-vstis/3189/3) where the same fix was probosed before.


ed95:
Added in [e7dbf60](https://github.com/WeAreROLI/JUCE/commit/e7dbf60fed49155d4a9a084c18333da61c4b44c8) - thanks for reporting!

PR #615, opened 2019-10-16
Update the AU host parameter cache when kAudioUnitProperty_ParameterList event is received

seb-nektar:
Some plugins (e.g Kontakt) dynamically change their parameters during runtime. The JUCE AU host internally holds a cached list of plugin parameters that has to be updated/refreshed when the plugin sends a kAudioUnitProperty_ParameterList event.

### DIFF
--- a/modules/juce_audio_processors/format_types/juce_AudioUnitPluginFormat.mm
+++ b/modules/juce_audio_processors/format_types/juce_AudioUnitPluginFormat.mm
@@ -1791,7 +1791,10 @@ private:
 
             default:
                 if (event.mArgument.mProperty.mPropertyID == kAudioUnitProperty_ParameterList)
+                {
+                    refreshParameterList();
                     updateHostDisplay();
+                }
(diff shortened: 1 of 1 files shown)


ed95:
Added in [0da9ed1](https://github.com/WeAreROLI/JUCE/commit/0da9ed11dc347bb182c7ab5d1824d0d8bcae9cda) - thanks for reporting!

ed95:
Actually, after some further testing I'm going to have to revert this. Although it looks simple enough to just update the cache, JUCE's plug-in wrapper and hosting classes aren't really capable of supporting dynamic parameters. In this case, the change would invalidate all the parameter listeners attached to the plug-in and would be a silently breaking (and crashing!) change.

seb-nektar:
Ok, I see. We did not encounter any problems with the added `refreshParameterList()`. But that might be because we do not use the JUCE parameter GUI classes. So unfortunately it seems there is no easy general fix and I will close this PR.

PR #622, opened 2019-10-22
Fixed ADSR hanging notes

maxwellpollack:
Currently, notes released before decaying to the sustain level have longer release times (inversely proportional to the sustain level), which is strange behavior for an ADSR envelope. This fix ensures that all notes have the same release times.

[ADSR sustain and release stage not working properly](https://forum.juce.com/t/adsr-sustain-and-release-stage-not-working-properly/32978)

[Bug in ADSR?](https://forum.juce.com/t/bug-in-adsr/32157)


ed95:
Added in [91513f5](https://github.com/WeAreROLI/JUCE/commit/91513f579db201ae3397a8b1109c663067361496), thanks!

PR #638, opened 2019-11-21
ZipFile : Add support for Zip64 (Compressed file over 4GB)

flyingrub:


jrlanglois:
Wouldn't this risk breaking in real-time under 32-bit apps and/or 32-bit OSs?

flyingrub:
Zip64 is a file format and not related to your os or app. I am not sure how JUCE handle those thing but for what I know 32 bit app / OS is just slower.

PR #640, opened 2019-12-03
Fix/bela fixes

giuliomoro:
A couple of fixes for the Bela backend

@cesaref


julianstorer:
Cheers Guilio, these changes should be on develop now

PR #641, opened 2019-12-04
Add option to use 'real' assert

abrykt:


modules/juce_core/system/juce_PlatformDefs.h
```diff
@@ -20,6 +20,13 @@
   ==============================================================================
 */
 
+#if JUCE_USE_CASSERT_ASSERT && JUCE_DEBUG
+#include <cassert>
+#define JUCE_CASSERT_FALSE assert(false)
```
jrlanglois, line 25: IMO, this should be moved closer to the `jassertfalse` implementation, and have `assert (false)` wrapped with `JUCE_BLOCK_WITH_FORCED_SEMICOLON`.

PR #642, opened 2019-12-04
Fix config flag to match actual define

McMartin:
@ed95 @tpoole something I found when working on `FindJUCE.cmake`.


ed95:
Added in [0273a4a](https://github.com/WeAreROLI/JUCE/commit/0273a4ad184db10eb620f5404e481e26f333980c), thanks @McMartin

PR #650, opened 2019-12-29
Un-constify getProgramName return value

danra:
`getProgramName` unconventionally returns `const String` rather than just `String`, and so triggers various warnings. Besides having no advantage AFAICS*, `clang-tidy` triggers the `readability-const-return-type` warning on this.

This is a breaking change for existing overriders, though (*which is a valid reason to keep as-is, though not a great one), so if the pull request is accepted it should also be listed under the breaking changes.


danra:
Too old and not important enough. Closing

PR #653, opened 2020-01-06
Fix some typos in juce_core

McMartin:
I got inspired by https://github.com/WeAreROLI/JUCE/pull/652.

PR #654, opened 2020-01-06
Fix some typos

McMartin:
Follow-up of #653.


ed95:
Added in [a54da0b](https://github.com/WeAreROLI/JUCE/commit/a54da0b832681c8567a9b5046e423fce3195b19a), thanks again! My personal favourite was destiuish -> distinguish..

What tool are you using for finding these? Would probably be a good idea for us to run it on the repo every now and then.

McMartin:
You're welcome! 
I did it quite manually, using the spell checking built-in feature of Sublime Text. So unfortunately, no script that you could run on the whole repo.

PR #656, opened 2020-01-13
add an internal MIDI out node

ShigekiKarita:
I love the AudioPluginHost to test plugins but it lacks of a midi output node. So I added it.


ed95:
There's a bit more plumbing involved in addition to adding the node, but we've added this in [4b118bb](https://github.com/WeAreROLI/JUCE/commit/4b118bb45b1c2c948b1a57daebf86f6b2a4c871c).

PR #658, opened 2020-01-18
Always include JuceHeader.h from the include search path

McMartin:
There are many files in JUCE that contain `#include "../JuceLibraryCode/JuceHeader.h"`, though there is no such `JuceHeader.h` file at that relative location. For instance:

https://github.com/WeAreROLI/JUCE/blob/724ae27c71ef809b836df06713cf9e50bc14162b/examples/DemoRunner/Source/Demos/JUCEDemos.cpp#L27
but [`JUCE/examples/DemoRunner/Source/JuceLibraryCode/JuceHeader.h` doesn't exist](https://github.com/WeAreROLI/JUCE/blob/724ae27c71ef809b836df06713cf9e50bc14162b/examples/DemoRunner/Source/JuceLibraryCode/JuceHeader.h).

The code compiles because the `JuceLibraryCode` folder itself is part of the include search path. In more details:
- the compiler first fails to find `JuceHeader.h` relatively to the source file that includes it,
- it falls back to the include search path,
- it resolves `../JuceLibraryCode/JuceHeader.h` from `JuceLibraryCode`, which always works for any project generated by Projucer.

Since JUCE 4.2.0, Projucer has always added the `JuceLibraryCode` folder to the include search path, so it is safe to write `#include <JuceHeader.h>` in any project generated by Projucer.

Thus I propose to always use `#include <JuceHeader.h>`. It is shorter, it stands out from the other relative includes, and it is always correct. Also, it simplifies a bit the code of Projucer, which doesn't have to deal with computing relative paths to `JuceHeader.h` anymore.

@ed95 @tpoole feel free to merge at will!


ed95:
Added in [7d330ab](https://github.com/WeAreROLI/JUCE/commit/7d330ab78241e05051404debb65dac2c9a5d5839) and [7484ac3](https://github.com/WeAreROLI/JUCE/commit/7484ac3dd3166040cef5e1bf26dd1db6841f6532). Thanks @McMartin !

McMartin:
@ed95 thank you for integrating these changes!

PR #659, opened 2020-01-30
Make NetworkServiceDiscovery::Advertiser::sendBroadcast() broadcast on all NICs

TSlivede:
I couldn't get `NetworkServiceDiscovery` to work when the device was only accessible via the second ethernet card.

I found, that `NetworkServiceDiscovery::Advertiser::sendBroadcast()` only broadcasts to the first ethernet card. That's hardcoded by [using `IPAddress::getLocalAddress()`](https://github.com/WeAreROLI/JUCE/blob/develop/modules/juce_events/interprocess/juce_NetworkServiceDiscovery.cpp#L70), which by the [description](https://docs.juce.com/master/classIPAddress.html#a1c01e16de47b2969ffbcf072d683c72c) only returns "the first 'real' address for the local machine".

The proposed change makes `NetworkServiceDiscovery::Advertiser::sendBroadcast()` iterate over all available local addresses and sends a broadcastmessage on each interface.

This solves the issue for me.

### DIFF
--- a/modules/juce_events/interprocess/juce_NetworkServiceDiscovery.cpp
+++ b/modules/juce_events/interprocess/juce_NetworkServiceDiscovery.cpp
@@ -67,11 +67,13 @@ void NetworkServiceDiscovery::Advertiser::run()
 
 void NetworkServiceDiscovery::Advertiser::sendBroadcast()
 {
-    auto localAddress = IPAddress::getLocalAddress();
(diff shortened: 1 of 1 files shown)


ed95:
Added in [c853277](https://github.com/WeAreROLI/JUCE/commit/c8532775e52c81a2738f8f239e9240d127ae7eb4). Thanks!

Issue #614, opened 2019-10-16
Korg Gadgets VST plugin UI vanishes on OSX


seb-nektar:
Steps to reproduce:

1. Load any of the Korg Gadget VST plugins in a JUCE-based plugin host
2. Open the plugin UI window
3. The plugin UI is visible for a split second and then vanishes

This was also reported in the [JUCE forum](https://forum.juce.com/t/korg-gadgets-plugin-editor-doesnt-work-with-juce-hosts/30824).

We debugged this issue a bit further and could find some indications what causes the problem:

VST2 plugin UI creation starts from `juce_VSTPluginFormat.cpp` and the instantiation of a [`VSTPluginWindow`](https://github.com/WeAreROLI/JUCE/blob/558ca3e5e6a731513de56c0cf3ad9120db70b449/modules/juce_audio_processors/format_types/juce_VSTPluginFormat.cpp#L2766). This contains a [`AutoResizingNSViewComponentWithParent` object](https://github.com/WeAreROLI/JUCE/blob/558ca3e5e6a731513de56c0cf3ad9120db70b449/modules/juce_audio_processors/juce_audio_processors.cpp#L98) which itself has a timer that resets the child view 30ms after creation. The timeout is exactly the time it takes until the KORG plugin UI vanishes. Removing the timer resolves the problem. But I suspect the timer is there for a reason...

Looking into the related code with git blame showed that the `AutoResizingNSViewComponentWithParent´ helper class was modified in commit 7c0d0ae. Most notably, the explicit setting of the child size has been removed. Maybe that could also be related to the problem we encounter with the KORG plugin.

mfisher31:
I have several reports of this exact same thing in my JUCE based hosts.

PR #663, opened 2020-02-05
Restructure AutoResizingNSViewComponentWithParent to maintain a valid parent pointer during its lifetime

seb-nektar:
This PR proposes a change to `AutoResizingNSViewComponentWithParent` such that its parent `NSView*` is not changed when instantiating VST2/3 plugin editors.

It fixes a crash when showing Softube plugin GUIs in JUCE hosts as well as visibility issues with KORG Gadget plugins (see #614). Both issues can be tested in the AudioPluginHostDemo.

The PR is marked as draft because it still contains an ugly workaround to trigger a resize of the parent window after the view has been attached in `AutoResizingNSViewComponentWithParent::timerCallback()`. Maybe one of the JUCE devs has an idea how to directly trigger the respective events when adding the `NSView` child.


seb-nektar:
This was finally fixed by commit 0f690662d8.

PR #668, opened 2020-02-11
First pass support for MSVC ARM64

jwinarske:
I have a first pass on support for MSVC ARM64.  Less Native SIMD support; as the code need some work.  MSFT ARM64 compiler doesn't like the subscript access without type  being defined as such.

Target tested on:  Surface ProX
Projucer runs fine.  DemoRunner launches, and the Demo List does not populate.

MSVC ARM64 supports OpenGL via ANGLE (OpenGL32 -> DirectX shim)
https://chromium.googlesource.com/angle/angle/+/refs/heads/master

Using FRUT:
c:\dev\repos\FRUT\prefix\FRUT\bin\Jucer2Reprojucer.exe Projucer.jucer c:\dev\repos\FRUT\cmake\Reprojucer.cmake
cmake .. -G"Visual Studio 16 2019" -AARM64

### DIFF
--- a/modules/juce_audio_basics/buffers/juce_FloatVectorOperations.cpp
+++ b/modules/juce_audio_basics/buffers/juce_FloatVectorOperations.cpp
@@ -1033,9 +1033,11 @@ intptr_t JUCE_CALLTYPE FloatVectorOperations::getFpStatusRegister() noexcept
     intptr_t fpsr = 0;
   #if JUCE_INTEL && JUCE_USE_SSE_INTRINSICS
     fpsr = static_cast<intptr_t> (_mm_getcsr());
-  #elif defined (__arm64__) || defined (__aarch64__) || JUCE_USE_ARM_NEON
+  #elif defined (__arm64__) || defined (__aarch64__) || defined (_M_ARM64) || JUCE_USE_ARM_NEON
    #if defined (__arm64__) || defined (__aarch64__)
     asm volatile("mrs %0, fpcr" : "=r" (fpsr));
+   #elif defined (_M_ARM64)
+    fpsr = _ReadStatusReg(ARM64_FPSR);
    #elif JUCE_USE_ARM_NEON
     asm volatile("vmrs %0, fpscr" : "=r" (fpsr));
    #endif
@@ -1053,9 +1055,11 @@ void JUCE_CALLTYPE FloatVectorOperations::setFpStatusRegister (intptr_t fpsr) no
   #if JUCE_INTEL && JUCE_USE_SSE_INTRINSICS
     auto fpsr_w = static_cast<uint32_t> (fpsr);
     _mm_setcsr (fpsr_w);
-  #elif defined (__arm64__) || defined (__aarch64__) || JUCE_USE_ARM_NEON
+  #elif defined (__arm64__) || defined (__aarch64__) || defined (_M_ARM64) || JUCE_USE_ARM_NEON
    #if defined (__arm64__) || defined (__aarch64__)
     asm volatile("msr fpcr, %0" : : "ri" (fpsr));
+   #elif defined (_M_ARM64)
+    _WriteStatusReg(ARM64_FPSR, fpsr);
    #elif JUCE_USE_ARM_NEON
     asm volatile("vmsr fpscr, %0" : : "ri" (fpsr));
    #endif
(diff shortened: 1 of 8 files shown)


modules/juce_dsp/juce_dsp.cpp
```diff
@@ -74,7 +74,7 @@
   #else
    #include "native/juce_sse_SIMDNativeOps.cpp"
   #endif
- #elif defined(__arm__) || defined(_M_ARM) || defined (__arm64__) || defined (__aarch64__)
+ #elif defined(__arm__) || defined(_M_ARM) || defined (__arm64__) || defined (__aarch64__) || defined(_M_ARM64)
```
jrlanglois, line 77: This is a bit strange; I wonder why this LOC isn't just `#elif JUCE_ARM` ?

modules/juce_core/native/juce_win32_SystemStats.cpp
```diff
@@ -166,6 +168,30 @@ void CPUInformation::initialise() noexcept
     hasAVX512VL        = (info[1] & (1u << 31)) != 0;
     hasAVX512VBMI      = (info[2] & (1u <<  1)) != 0;
     hasAVX512VPOPCNTDQ = (info[2] & (1u << 14)) != 0;
+#else
+    hasMMX = FALSE;
```
jrlanglois, line 172: Not sure why you're sticking the Win32 `FALSE` here instead of the C++ `false`? Regardless, those values will always be `false`; ARM doesn't have SSE or AVX registers.

jwinarske: Yeah it's a bit of a bolt-on.  I think there could be some effort put into avoiding the interleaved target conditionals.  I see some arguments for having a separate file per platform.

modules/juce_core/native/juce_win32_SystemStats.cpp
```diff
@@ -60,7 +60,8 @@ static void callCPUID (int result[4], uint32 type)
 #else
 static void callCPUID (int result[4], int infoType)
 {
-   #if JUCE_PROJUCER_LIVE_BUILD
+   #if JUCE_PROJUCER_LIVE_BUILD || defined(_M_ARM) || defined(_M_ARM64)
+    (void)infoType;
```
jrlanglois, line 64: The usual JUCE way is to make use of [`ignoreUnused`](https://github.com/juce-framework/JUCE/blob/develop/modules/juce_core/maths/juce_MathsFunctions.h#L286-L289) in these situations: `ignoreUnused (infoType);`

modules/juce_audio_basics/buffers/juce_FloatVectorOperations.cpp
```diff
@@ -1033,9 +1033,11 @@ intptr_t JUCE_CALLTYPE FloatVectorOperations::getFpStatusRegister() noexcept
     intptr_t fpsr = 0;
   #if JUCE_INTEL && JUCE_USE_SSE_INTRINSICS
     fpsr = static_cast<intptr_t> (_mm_getcsr());
-  #elif defined (__arm64__) || defined (__aarch64__) || JUCE_USE_ARM_NEON
+  #elif defined (__arm64__) || defined (__aarch64__) || defined (_M_ARM64) || JUCE_USE_ARM_NEON
```
jrlanglois, line 1036: The `_M_ARM64` is redundant because you're defining [`JUCE_USE_ARM_NEON`](https://github.com/juce-framework/JUCE/pull/668/files#diff-49c75ab597f7147dafdc5fbe6abe4580R90).

modules/juce_audio_basics/buffers/juce_FloatVectorOperations.cpp
```diff
@@ -1053,9 +1055,11 @@ void JUCE_CALLTYPE FloatVectorOperations::setFpStatusRegister (intptr_t fpsr) no
   #if JUCE_INTEL && JUCE_USE_SSE_INTRINSICS
     auto fpsr_w = static_cast<uint32_t> (fpsr);
     _mm_setcsr (fpsr_w);
-  #elif defined (__arm64__) || defined (__aarch64__) || JUCE_USE_ARM_NEON
+  #elif defined (__arm64__) || defined (__aarch64__) || defined (_M_ARM64) || JUCE_USE_ARM_NEON
```
jrlanglois, line 1058: Same here, and the various areas below.

modules/juce_core/native/juce_win32_SystemStats.cpp
jrlanglois: Bolt-on to achieve... what? I'm not sure I understand here.

adiantek:
When it will be merged? It's needed by OpenShot.

PR #646, opened 2019-12-13
AU, improved support for kAudioUnitProperty_PresentPreset

seb-nektar:
This PR consists of two parts:

1. Call `updateHostDisplay()` when `kAudioUnitProperty_PresentPreset` has changed

2. Return `kAudioUnitProperty_PresentPreset` name when querying `getProgramName(-1)`

Many **AU** plugins report their currently active preset name (`kAudioUnitProperty_PresentPreset`) although they have an empty factory program list (`kAudioUnitProperty_FactoryPresets`).  At the same time they consistently report a current program index of -1.

These are for example: 

- U-He products
- Novation V-station + BassStation
- Audjoo Helix
- KV331 Synthmaster products
- Sonic Charge products

It would be nice if JUCE hosts could retrieve the name of `kAudioUnitProperty_PresentPreset` by querying the -1 program index.

PR #670, opened 2020-02-26
Fix vst program access

seb-nektar:
This PR contains two fixes to VST2 and VST3 program handling.

- First part fixes a crash when trying to retrieve the current program index from a VST3 plugin that does not have a program list, e.g. Slate Virtual Mix Rack or Eiosis AirEQ

- Second part makes sure an empty string is returned when the supplied index is invalid  for VST2 and VST3

One could argue that the caller should check `getNumPrograms()` before accessing any of these methods. However, some AU plugins report 0 programs but can still report the current program name at index -1 (see #646). So from the hosting code it is difficult to filter valid program indexes without adding branches for the various plugin formats. IMO it makes more sense to simply make the respective plugin wrapper code more roubust instead.

### DIFF
--- a/modules/juce_audio_processors/format_types/juce_VST3PluginFormat.cpp
+++ b/modules/juce_audio_processors/format_types/juce_VST3PluginFormat.cpp
@@ -2239,8 +2239,15 @@ public:
 
     //==============================================================================
     int getNumPrograms() override                        { return programNames.size(); }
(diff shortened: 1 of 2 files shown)


ed95:
Added in [f3470d3](https://github.com/WeAreROLI/JUCE/commit/f3470d3cd45e39066e4b048b5d036e7996713c74). Thank you for reporting!

PR #671, opened 2020-03-04
Fix debugger breakpoint triggering on Linux

a1exwang:
On Linux, when using `jassert` under gdb, gdb won't correctly break on the line of `jassert`.
Since `::kill` is asynchronous, it the program may not receive the signal before kill returns. In that case, gdb may break on some random point, which will causes frustration.

We can replace kill() with abort(), which is synchronous, according POSIX standard:

from https://pubs.opengroup.org/onlinepubs/9699919799/functions/abort.html
>> The SIGABRT signal shall be sent to the calling process as if by means of raise() with the argument SIGABRT.

from https://pubs.opengroup.org/onlinepubs/9699919799/functions/raise.html
>> The raise() function shall send the signal sig to the executing thread or process. If a signal handler is called, the raise() function shall not return until after the signal handler does.


modules/juce_core/system/juce_PlatformDefs.h
```diff
@@ -57,13 +57,16 @@ namespace juce
...
+#if JUCE_IOS
   /** This will try to break into the debugger if the app is currently being debugged.
       If called by an app that's not being debugged, the behaviour isn't defined - it may
       crash or not, depending on the platform.
       @see jassert()
   */
   #define JUCE_BREAK_IN_DEBUGGER        { ::kill (0, SIGTRAP); }
+#elif JUCE_LINUX
+  #include <cstdlib>
+  #define JUCE_BREAK_IN_DEBUGGER        { std::abort(); }
```
McMartin: As noted in the pull request description, `std::abort()` will result in `raise(SIGABRT)`, which will force the program to end. It would be better to use `raise(SIGTRAP)` here.

a1exwang: It makes sense.

a1exwang: Changed and tested on Arch Linux latest version

PR #675, opened 2020-03-13
Avoid passing nullptr to setBusArrangements()

seb-nektar:
This applies the same fix from commit 5a0d4098 to the second occurence of `setBusArrangements()` a few lines later in `syncBusLayouts()`.

Without the patch we encountered crashes when loading the Aly James VST3 plugins [VSDX ](https://www.alyjameslab.com/alyjameslabvsdsx.html) or [VPRom](https://www.alyjameslab.com/alyjameslabvlinn.html). Applying the patch fixes the crash.


ed95:
Thanks! Added in [9f0728e](https://github.com/WeAreROLI/JUCE/commit/9f0728e6226174b7e933bff1123a2aca1f2c9a33).

PR #676, opened 2020-03-13
Check if parameter list is empty when setting haveParameterList flag

seb-nektar:
The parameter list of Aly James **AU** plugins [VSDX ](https://www.alyjameslab.com/alyjameslabvsdsx.html) or [Claptrap ](https://www.alyjameslab.com/alyjameslabclaptrap.html) is empty when hosted in JUCE.

These plugins do not report an error when querying `kAudioUnitProperty_ParameterList` but return a `paramListSize` of zero when refreshing the parameter list early in the init sequence. Hence, `haveParameterList`  should be set to false. Later on in [prepareToPlay(), line 989](https://github.com/seb-nektar/JUCE/blob/ce5981c80a3e635479bd2fd7a027c5e4bc20663a/modules/juce_audio_processors/format_types/juce_AudioUnitPluginFormat.mm#L989) the parameter list will be queried again if `haveParameterList` is false. Now the plugins return a non-empty parameter list.

### DIFF
--- a/modules/juce_audio_processors/format_types/juce_AudioUnitPluginFormat.mm
+++ b/modules/juce_audio_processors/format_types/juce_AudioUnitPluginFormat.mm
@@ -1401,8 +1401,10 @@ public:
         if (audioUnit != nullptr)
         {
             UInt32 paramListSize = 0;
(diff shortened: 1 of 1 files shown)


ed95:
Added in [375c746](https://github.com/WeAreROLI/JUCE/commit/375c7462515f26574436c7806b24402be60ad0f5). Thank you for reporting!

PR #683, opened 2020-03-23
Wrong comments in KeyPress::createFromDescription().

Dicklessgreat:
"SPACE" didn't work. 
It should be "spacebar".
And I don't know it's good comment or not, but I had to "@see" KeyPressHelpers::translations.
So I added that at last.

### DIFF
--- a/modules/juce_gui_basics/keyboard/juce_KeyPress.h
+++ b/modules/juce_gui_basics/keyboard/juce_KeyPress.h
@@ -135,14 +135,14 @@ public:
     //==============================================================================
     /** Converts a textual key description to a KeyPress.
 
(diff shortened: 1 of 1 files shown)


ed95:
Thanks for reporting. We've updated the incorrect docs in [5315160](https://github.com/WeAreROLI/JUCE/commit/5315160e5119ed17c26f8a3758cdd4a902cbc18e) but I'm not sure pointing people towards `KeyPressHelpers::translations` is a good idea. As per the docs for `createFromDescription()` you shouldn't be passing your own strings to that method directly - 

"This isn't designed to cope with any kind of input, but should be given the strings that are created by the getTextDescription() method."

PR #695, opened 2020-04-17
Convert juce::String to utf16 before assigning it to Steinberg::UString

seb-nektar:
`Steinberg::UString` internally uses an utf16 encoding. `Steinberg::UString::fromAscii` expects a pure ASCII string and simply casts each character to a utf16 character but cannot convert utf8 encoded strings. 

With this patch the function first converts the `juce::String` to a utf16 string and then assigns it to the `Steinberg::UString` to maintain correct string encoding.


seb-nektar:
The first commit did create a compile error with XCode. The second commit fixes this by using an `reinterpret_cast<>` and the already existing `toString()` function.

seb-nektar:
This was fixed by e1f892f.

PR #696, opened 2020-04-18
Speed-up building Projucer with CMake

McMartin:
@reuk there might be other flags on other targets to add in a similar way, but I usually only build Projucer.


McMartin:
Made obsolete by https://github.com/juce-framework/JUCE/commit/eb01832c489f7a8e5abb7d23ff4640ef28fe29c0.

PR #697, opened 2020-04-18
Make OggWriter::OggWriter() safely callable from multiple threads

danra:


danra:
@reuk This one still isn't resolved. Can be fixed with my very old suggested patch or using a similar approach to the `LockedRandom` you used in 093dbc7

Anthony-Nicholls:
@danra just keep you posted I've looked into this issue and agree that using `getSystemRandom()` isn't the right thing to do here. I'm looking at removing calls to `getSystemRandom()` from all JUCE library code, hopefully we'll have something merged soon (likely next week now).

Thanks for reporting.

Anthony-Nicholls:
@danra I know this issue is marked as closed but I thought it worth adding on here that getSystemRandom() should now be thread-safe by default https://github.com/juce-framework/JUCE/commit/8fe8717ebc9ba0f3c6632f86b17044d4c936b338

danra:
@Anthony-Nicholls Great, thanks for letting me know
p.s. Perhaps worth mentioning under breaking changes?

PR #698, opened 2020-04-18
Make TemporaryFile ctor safely callable from multiple threads

danra:


kunitoki:
maybe it should be better to make getSystemRandom threadsafe instead? a default constructed random with the same seed will always return the same value?

danra:
Was fixed in the meantime. Closing.

PR #700, opened 2020-04-19
fix cast-qual warnings that showed up when compiling with clang-cl

TobiasLudwig:
* in most cases the c-style casts were just missing the `const`, so the fix was straightforward
* a non-const overload for `MidiMessage::getData` was added to fix const-correctness 
* the warning was explicitly disabled for third party code that has already warnings disabled (e.g. flac, OggVorbis, zlib)
* the signature of `AudioFormatWriter::write` was changed to elide const casts (`const int**` --> `const int* const*`)
* the signature of `ConfigMetaData::ConfigMetaData` was changed to elide const casts and to correctly detect the array size
* a member of `CriticalSection` was made `mutable` to elide const casts. This fits the non-Windows implementation where the member is already `mutable`
* in some cases casting away the constness was actually necessary for calling windows api functions so explicit `const_cast`s where added to make this clear (`juce_win32_Midi.cpp` and `juce_win32_Network.cpp`)

### DIFF
--- a/modules/juce_audio_basics/midi/juce_MidiMessage.cpp
+++ b/modules/juce_audio_basics/midi/juce_MidiMessage.cpp
@@ -109,7 +109,7 @@ MidiMessage::MidiMessage (const void* const d, const int dataSize, const double 
 {
     jassert (dataSize > 0);
     // this checks that the length matches the data..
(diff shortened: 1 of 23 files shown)


tpoole:

PR #706, opened 2020-04-25
Better environment control

kunitoki:
Add support for better environment variables in the framework, because since today it was not possible to start a process and control which environment should have:

- Added functions to `SystemStats` to allow getting and changing the environment:
```
bool setEnvironmentVariable(const String& name, const String& value)

bool removeEnvironmentVariable(const String& name)

StringPairArray getEnvironmentVariables()
```

- Added functions to `ChildProcess` to pass the environment variables the child process should retain:
```
bool start (const String& command, const StringPairArray& environment, int streamFlags = wantStdOut | wantStdErr)
bool start (const StringArray& arguments, const StringPairArray& environment, int streamFlags = wantStdOut | wantStdErr)
```

- Added a function to `Process` to pass the environment variables the process spawned should retain:
```
static bool JUCE_CALLTYPE openDocument (
    const String& documentURL,
    const String& parameters,
    const StringPairArray& environment)
```


tpoole:

PR #707, opened 2020-04-25
Fix URL issue removing get parameters and adding them to post data

kunitoki:
When you have post data while using `createInputStream`, JUCE will end up removing the get parameters from the endpoint, and append them to the body, breaking both the endpoint to reach and the payload.

This also fix `createInputStream` to take a `std::function` for the progress callback instead of a function pointer and void* context.


tpoole:

JordanSkousen:
Reopened this PR on master branch, see #830

PR #709, opened 2020-04-26
Fixes: Undefined symbol juce_iOSMain for standalone iOS application

phniix:
Fixes: Implements the proposed solution in juce-framework/JUCE/#708

Adds `JUCE_MODULE_AVAILABLE_juce_gui_basics` check to ensure that no linker errors occur if the `juce_gui_basics` module is not included in a standalone iOS application.


ed95:
Thanks for the report and fix. We've added this in https://github.com/juce-framework/JUCE/commit/b57509c4a21bafa54341a085ef48683cf16039e5

PR #710, opened 2020-05-01
Fix crash on quit in WaveLab 10 on OS X.

Pflugshaupt:
A crash happens if multiple instances of the same JUCE vst3 plugins are active and WaveLab 10 is quit.

Lots of details in my monologue on the forum:

https://forum.juce.com/t/wavelab-10-crash-on-quit-inside-vst3-wrapper/39067/5


devinkerr:
Can someone at JUCE take a look at this?

PR #722, opened 2020-05-14
Fix hidpi scaling on Linux

terencode:
@falkTX and I were trying to figure out why a VST (Pianoteq) was not displaying correctly on Carla as well as Reaper or even jalv (https://gitlab.com/drobilla/jalv) on my setup (https://github.com/falkTX/Carla/issues/932)  and we found the problem:

I'm on Xorg with gnome-shell and a hidpi screen (2560x1440) so  `getDisplayScale()` from https://github.com/juce-framework/JUCE/blob/juce6/modules/juce_gui_basics/native/x11/juce_linux_XWindowSystem.cpp#L909 returns 2. JUCE then scales the plugin accordingly but the problem is it's not notifying the host about it. 
So, what happened was only 1/4 of a plugin UI would be shown.
I could reproduce this with Pianoteq 6 and CHOW (https://github.com/jatinchowdhury18/CHOW) but it probably impacts all the plugins built with JUCE5/6

Once those commits are applied, the VST2/3 will need to be rebuilt and will in turn correctly work on hidpi (we tested it and confirmed working with CHOW).

### DIFF
--- a/modules/juce_audio_plugin_client/VST/juce_VST_Wrapper.cpp
+++ b/modules/juce_audio_plugin_client/VST/juce_VST_Wrapper.cpp
@@ -913,7 +913,7 @@ public:
             if (auto* ed = processor->createEditorIfNeeded())
             {
                 setHasEditorFlag (true);
-                editorComp.reset (new EditorCompWrapper (*this, *ed));
+                editorComp.reset (new EditorCompWrapper (*this, *ed, lastScaleFactorReceived));
             }
             else
             {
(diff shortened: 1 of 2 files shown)


terencode:
Some screenshots of the problem: 

In Reaper:
 
image

In Carla:
image

julienpommier:
It is quite easy to reproduce with the JUCE AudioPluginDemo VST3, just build it with getDisplayScale returning 2 and load it in the AudioPluginHost.

ed95:
Thanks for the report! The approach that we took for Windows high DPI scaling in plug-ins was to bypass the normal scaling that happens in the `ComponentPeer` code for standalone applications and instead let the host control the scale factor of the plug-in. We've added something similar to the Linux code in [a8b6066](https://github.com/juce-framework/JUCE/commit/a8b606618767a8696ae0793a27cd7a9db896a812) which should fix the issue that you're seeing.

PR #727, opened 2020-05-21
Remove unused function

McMartin:
It was introduced in 14d7675c48b54c5ecf6d45f6fa77d73079ca635c, but was not used.

@ed95 this should be an easy one.


ed95:
Added here https://github.com/juce-framework/JUCE/commit/257e5e579b25d2908c41bf72fa3acc1850688b53 - thanks!

PR #732, opened 2020-05-27
Correctly handle aftertouch and channel pressure in VST3 plugin code

seb-nektar:
This patch fixes aftertouch and channel pressure handling in VST3 plugin code.

Note that JUCE and VST3 SDK terminology is different and misleading. What is called aftertouch in  `juce::MidiMessage` corresponds to VST3 `kPolyPressurerEvents`. Channel pressure from a `juce::MidiMessage` should be mapped to the `kAfterTouch` parameter via MIDI controller assignments.

Also see this Steinberg dev forum thread: https://sdk.steinberg.net/viewtopic.php?f=4&t=730

### DIFF
--- a/modules/juce_audio_processors/format_types/juce_VST3Common.h
+++ b/modules/juce_audio_processors/format_types/juce_VST3Common.h
@@ -501,8 +501,9 @@ public:
                         break;
 
                     case Steinberg::Vst::Event::kPolyPressureEvent:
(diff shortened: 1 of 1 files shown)


seb-nektar:
Obsolete since changes applied in 4a613dfad2a90bc1fd738990127ef071aad28a53.

PR #736, opened 2020-06-08
Adding MouseInterceptor class for better AAX/VST3 menu handling

LukeSonnox:
These changes are an example of how to implement a class which allows juce::MouseEvent calls to be intercepted and stolen by an external object.

- This allows for the AAX wrapper to implement the expected behaviour when showing the parameter automation menu (i.e. mouseDown and mouseDrag events shouldn't be allowed to the target juce::Component when the keys are pressed);

- Added a workaround for the missing AAX_eModifiers_SecondaryButton flag from GetModifiers function;

- Added the parameter context menu to the VST3 wrapper using the same mechanism.

A couple of things:
- There is a lot of duplication between the MouseListenerList and MouseInterceptorList classes which can probably be reduced;
- I haven't compiled and tested the g++ 4.8 checks;
- I've only implemented mouseDown, mouseDrag and mouseUp in this example because that's all that's needed for the parameter context menus, but we have examples where it's convienient for other mouse function calls to be intercepted for various reasons.

### DIFF
--- a/modules/juce_audio_plugin_client/AAX/juce_AAX_Wrapper.cpp
+++ b/modules/juce_audio_plugin_client/AAX/juce_AAX_Wrapper.cpp
@@ -568,6 +568,7 @@ namespace AAXClasses
 
         //==============================================================================
         struct ContentWrapperComponent  : public Component
+                                        , public MouseInterceptor
         {
             ContentWrapperComponent (JuceAAX_GUI& gui, AudioProcessor& plugin)
                 : owner (gui)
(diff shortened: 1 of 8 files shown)


tpoole:

LukeSonnox:
> The removal of the juce6 branch automatically closed this PR. Please re-submit against either `master` or `develop`. Apologies for the inconvenience.

I've reopened the pull request here: https://github.com/juce-framework/JUCE/pull/828

PR #747, opened 2020-06-30
GUI: FileBrowserComponent's go up button will now not disappear after a look and feel change

kamedin:
After being replaced in lookAndFeelChanged(), goUpButton has no parent, or tooltip, or onClick callback. As the constructor already calls lookAndFeelChanged(), these lines are moved.


kamedin:
Fixed [here](https://github.com/juce-framework/JUCE/commit/b096c16c042eef773b18f4fbe343abb3bf4c5276).

PR #754, opened 2020-07-08
Added conical gradient pattern.

ImJimmi:
### Added a conical gradient pattern to the `ColourGradient` class.

The conical gradient can be used by setting the `isConical` flag to true. The gradient will then be calculated using the angle clockwise from the start/end position (provided by `point2`) around a full 360 degrees back to the start/end position. `point1` is used as the centre of the cone.

### Example usage:
```
void MainComponent::paint (juce::Graphics& g)
{
    const auto centre = getLocalBounds().toFloat().getCentre();
    const auto pi = juce::MathConstants<float>::pi;

    juce::ColourGradient gradient (juce::Colours::purple, centre,
                                   juce::Colours::yellow, centre.withY (centre.y + 100.f),
                                   false, // We don't want a radial pattern
                                   true); // We DO want a conical pattern
    gradient.addColour (0.125, juce::Colours::purple);
    gradient.addColour (0.875, juce::Colours::yellow);
    g.setGradientFill (gradient);

    juce::Path p;
    p.addCentredArc (centre.x, centre.y, 100.f, 100.f, 0.f, -pi * 0.75f, pi * 0.75f, true);
    g.strokePath (p, juce::PathStrokeType (50.f));
}
```

> [Result of the above example.](https://i.imgur.com/jGFZyMa.png)

Setting both `isRadial` and `isConical` to true in the constructor will trigger an assertion warning that only one of the flags can be set to true at a time. If both flags are set to two after the object is constructed, the radial pattern takes presidence and the conical pattern will not be used.

The `isConical` argument added to the `ColourGradient` constructor is set to false by default so as to ensure backwards compatibility with existing code.

### Known issues:
- The closer `point2` is to `point1`, the less precise the gradient is which results in aliasing of the colours where distinct bands of colours can be seen. This can be avoided by ensuring the points are far enough apart (at least 50px seems to work well), however that may not be an obvious solution for some.
[Example where `point2` is just 3px below `point1`](https://i.imgur.com/PTcaUAS.png).

- The start/end boundary where the first and last colours meet is aliased which may be undesirable for many applications.
[Example showing aliasing where the colour wraps back around to the start colour](https://i.imgur.com/H5qEpOF.png).

### DIFF
--- a/modules/juce_graphics/colour/juce_ColourGradient.h
+++ b/modules/juce_graphics/colour/juce_ColourGradient.h
@@ -215,7 +215,7 @@ public:
     Point<float> point1, point2;
 
     /** Defines which of the available gradient types should be used. */
-    GradientType gradientType;
+    GradientType gradientType = GradientType::Linear;
 
     /** isRadial has been deprecated in favour of the gradientType property.
         See ColourGradient::GradientType.
--- a/modules/juce_graphics/colour/juce_ColourGradient.h
+++ b/modules/juce_graphics/colour/juce_ColourGradient.h
@@ -37,6 +37,16 @@ namespace juce
 class JUCE_API  ColourGradient  final
 {
 public:
+    //==============================================================================
+    /** The available gradient types. */
+    enum GradientType
+    {
+        Linear, // Colours transition along a straight line.
+        Radial, // Colours radiate from an origin.
+        Conical // Colours rotate around a centre point.
+    };
+
+    //==============================================================================
     /** Creates an uninitialised gradient.
 
         If you use this constructor instead of the other one, be sure to set all the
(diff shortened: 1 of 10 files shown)


modules/juce_graphics/colour/juce_ColourGradient.h
```diff
@@ -200,21 +214,13 @@ class JUCE_API  ColourGradient  final
...
-
-        If false and isConical is also false, the gradient is linear between the
-        two points.
-    */
-    bool isRadial;
-
-    /** If true, the gradient should be filled conically, centred around point1,
-        with point2 defining the start/end position.
+    /** Defines which of the available gradient types should be used. */
+    GradientType gradientType;
```
ffAudio, line 218: Instead of initialising in the constructor, how about using default initialisation? (available since C++11)
```
GradientType gradientType = Linear;
```

ImJimmi: Done!

modules/juce_graphics/colour/juce_ColourGradient.h
```diff
@@ -37,6 +37,16 @@ namespace juce
 class JUCE_API  ColourGradient  final
 {
 public:
+    //==============================================================================
+    /** The available gradient types. */
+    enum GradientType
```
jrlanglois, line 42: My suggestion here is to use an `enum class`.

In terms of naming conventions - as per the JUCE coding standards:
>For enums, use camel-case with the same capitalisation that you'd use for a class and its member variables, e.g.
```c++
enum class MyEnum
{
    enumValue1 = 0,
    enumValue2 = 1
};
```

jrlanglois:
This is pretty cool!

@ImJimmi Have you given this a shot in an OpenGL context?

Also, it's probably a good idea to extend a demo or create a PIP to put on here so people can give it a whirl. With that, it might be worthwhile having all 3 gradient types side by side to compare and be sure results make sense in a cross-platform rendering context (ie: iOS vs macOS vs Android vs Windows).

modules/juce_graphics/native/juce_RenderingHelpers.h
```diff
@@ -553,6 +553,46 @@ namespace GradientPixelIterators
...
+    //==============================================================================
+    /** Iterates the colour of pixels in a conical gradient */
+    struct Conical
+    {
+        Conical (const ColourGradient& gradient, const AffineTransform&,
+                 const PixelARGB* colours, int numColours)
+            : lookupTable (colours),
+              numEntries (numColours),
+              centre (gradient.point1),
+              offsetAngle (gradient.point1.getAngleToPoint(gradient.point2))
```
jrlanglois, line 566: I see some lacking space before `(` as per the JUCE coding standards in this new struct.

modules/juce_graphics/native/juce_RenderingHelpers.h
```diff
@@ -553,6 +553,46 @@ namespace GradientPixelIterators
...
+                angle += twoPi;
+
+            return lookupTable[jlimit(0, numEntries, roundToInt(numEntries * angle / twoPi))];
+        }
+
+        const PixelARGB* const lookupTable;
+        const int numEntries;
+        const Point<float> centre;
+        const float offsetAngle;
+        float y;
```
jrlanglois, line 591: `y` isn't initialised anywhere within the scope of this structure - might be worth setting it here.

kunitoki:
Thanks for the contribution, we really appreciate the feature and the work done to make this happen.

That said, the PR in this state can't be merged, there some small inconsistencies with the JUCE coding standard plus there are missing parts in the implementation that needs to be addressed before considering the feature production ready and mergeable.

The software rendered is not enough, JUCE supports other graphics rendering like OpenGL and CoreGraphics and those needs to be addressed too, or the feature will not be supported when using those renderers.
- [ ] Implement OpenGL support
- [ ] Implement CoreGraphics support
- [ ] While doing those implementations, it would be appreciated if we could extract the common code in a separate, testable helpers
- [ ] The helper code must be generic enough to be reused potentially by future rendering engine implementations in JUCE like vulkan or metal (sorry but at the moment we have not a clear roadmap ahead and we can't give much info when and how this will be done).

Also this does not contain a single unit test, and to be merged and consider it production ready we need at least to have some coverage (honestly, the `juce_graphics` module has very few unit tests at the moment limited to only graphics primitives like Colour and nothing else, but we will appreciate if you could start implementing some unit tests for the changes you did in the ColourGradient class).
- [ ] Implement unit tests

Finally, JUCE inconsistencies must be resolved.
- [ ] Fix JUCE coding standard inconsistencies

Once those missing parts will be done, we could consider cleaning up the feature for inclusion.

ed95:
Whilst @kunitoki has provided a good summary of the main issues with the PR in its current state and explained why it isn’t mergeable, this isn’t our official policy. We’re working on getting a proper PR checklist and CLA added to the repo so that these steps are clearer for contributors like yourself in the future, but in the meantime adding support to the other renderers in JUCE in a D.R.Y way is a good next step.

kunitoki:
Thank you!

PR #767, opened 2020-07-29
Write correct format of icon

McMartin:
@ed95 @reuk when looking at 450ac8ade557c60fe2f47ea862018118fc6d6c2f (_Projucer:  Don't assert when no icons are specified in Code::Blocks exporter_), I saw "icon.ico" next to "Mac", which didn't make sense. I found out that this was wrong since the beginning of the `juce6` branch.


reuk:
Thanks for bringing this to our attention, the change was applied in https://github.com/juce-framework/JUCE/commit/14eebfb85abb05c85fc482bb0bf99c7547c0052e

PR #775, opened 2020-09-08
Fixed a race condition affecting the JACK implementation

op414:
This fixes an issue where JUCE would not be able to correctly connect the
JACK ports of the JUCE app to the selected JACK ports of the selected
"device". The higher the channel count the more likely this race condition
will cause issues.


reuk:
This PR has been discussed on the JUCE forum, [here](https://forum.juce.com/t/jack-race-condition-pull-request/41658).

I believe that this issue is fixed in JUCE 6.

op414:
I can confirm that this is fixed in Juce 6.0.4

PR #779, opened 2020-09-23
Correctly forward MIDI channel pressure (0xd) as Steinberg::Vst::kAfterTouch event

seb-nektar:
Channel pressure MIDI CC messages (0xd) should be mapped as VST3 events of type `Steinberg::Vst::kAfterTouch`.


reuk:
Thanks, this should be resolved by https://github.com/juce-framework/JUCE/commit/90664b42df2771df0d5415b7e9b314b2eae8e73e

PR #785, opened 2020-10-02
Reference to repository containing compiled apps with log for legacy users (not merge)

AnonA2:
https://github.com/AnonA2/Juce3LegacyPullRequestCompiled/tree/master
OS X 10.7.5
See also log, there are some errors. IMO it always advised to provide at least basic support for OS X <10.7 and 10.8> separately. E.g. like mediahuman or other affiliates.

### DIFF
--- a/Changelog.txt
+++ b/Changelog.txt
@@ -1,0 +1,2 @@
+https://github.com/AnonA2/Juce3LegacyPullRequestCompiled/tree/master
+This contains compiled legacy apps for OS X 10.7.


tpoole:
I'm not sure what you are trying to achieve here.

The current distribution of JUCE allows you to create software that will run on macOS 10.7, but you need a more modern system (macOS 10.11 and Xcode 7.3.1) to compile this code.

If you only have access to a pre macOS 10.11 OS then you will need to use a much older version of JUCE if you want to use that machine as a development environment. It is not possible to use such an old setup to create software that will run on the latest Apple hardware (irrespective of the version of JUCE).

Even if you're happy to work with this restriction, providing precompiled apps for legacy distributions of JUCE won't help you - if you cannot compile the Projucer and the demo then you will not be able to compile any other JUCE-based software on that system.

From the contents of your PR it looks like you are trying to use JUCE 3. This was released 8 years ago and we no longer support it. You may be able to get some assistance from the helpful folk on the JUCE forum, https://forum.juce.com.

PR #787, opened 2020-10-08
Fix AUv3 indexed parameter automation issues

amethystdeceiver:
Parameter automation in AUv3 wrapper is missing a normalisation step, leading to a choice parameter reaching its maximum value at 1 instead of getMaximumParameterValue().


amethystdeceiver:
Closing, fixed in c492da9971dd3bdf59b9b62c94064a6e07f58252

PR #794, opened 2020-10-23
Pull request pace protected cmake fix

BastienC09:
SEE: https://forum.juce.com/t/cmake-ilok-garageband/42344


tpoole:
Added in ba172bc828317c33b9f1a7dca17b05bdd03ec84b.

PR #798, opened 2020-11-04
https://forum.juce.com/t/about-cmake-and-ios/42554

BastienC09:
SEE: https://forum.juce.com/t/about-cmake-and-ios/42554


BastienC09:
FOLLOW UP
https://github.com/juce-framework/JUCE/commit/4a036afe32600fe54f0f02c7651a9eb5de78c3f9

PR #800, opened 2020-11-13
Update juce_MemoryAudioSource.cpp

odecaux:
The clamping forces the sample to loop indefinitely, because the above condition (i < n || isCurrentlyLooping) is always satisfied.


reuk:
Thanks for raising this. The issue should be resolved by this commit: https://github.com/juce-framework/JUCE/commit/27ca5e6ccb69f9c1cf24c103935621bfe30195c8

PR #813, opened 2020-12-08
Mobile deep link

adamski:
Add support for "deep links" on iOS and Android. 

Following the existing pattern on desktop platforms, where a link from a website registered with an app will end up calling `JUCEApplication::anotherInstanceStarted` with the URL. 

On iOS it was necessary to replace the (now deprecated) `applicationDidFinishLaunching` with `applicationDidFinishLaunchingWithOptions`. 

I also added support for "Custom URL Schemes" on iOS via `openURL` but this is untested. 

On Android I have hijacked the `PushNotifications` class which deals with intents to check for a `VIEW` intent which has the `BROWSABLE` category. See https://developer.android.com/training/app-links/deep-linking


adamski:
Please note - on iOS this requires the "Associated Domains" capability to be set. So this should really be added to Projucer as part of the addition.

PR #816, opened 2020-12-15
 VST3 Program parameter is not updated when the loaded program changed and updateHostDisplay() is called

TheSlowGrowth:
Since commit #51fe471461f75fcfd72f7d0749382cbc149eef37 the VST3 Program parameter is not updated when the loaded program changed and updateHostDisplay() is called.


TheSlowGrowth:
Fixed in 3ff966603e57ba55c3ab372dd9ef2340114feed1

PR #823, opened 2021-01-05
Improvement to support simplest sequential animations

skident:
- added a delay before animation start which provides a possibility to make a simple sequential animation. Moreover it gives an opportunity to animate the second comoponent in the middle of the first component animation


modules/juce_gui_basics/layout/juce_ComponentAnimator.cpp
```diff
@@ -76,6 +77,8 @@ class ComponentAnimator::AnimationTask
                                        : component.get())
         {
             msElapsed += elapsed;
+            if (msElapsed < 0)
+                return true;
```
skident, line 81: msElapsed < 0 is okay if you there is a delay before animation start.

PR #830, opened 2021-01-14
[BUG] URL issue removing get parameters and adding them to post data

JordanSkousen:
This fixes a **bug** in JUCE where it removes URL parameters from the endpoint and appends them to the body when POST data is used.

This also changes `createInputStream` to take a `std::function` for the progress callback instead of a function pointer and void* context.

This is a re-creation of PR #707.

### DIFF
--- a/extras/Projucer/Source/Application/UserAccount/jucer_LicenseQueryThread.h
+++ b/extras/Projucer/Source/Application/UserAccount/jucer_LicenseQueryThread.h
@@ -322,7 +322,7 @@ private:
             return cancelledError;
 
         int statusCode = 0;
-        auto urlStream = url.createInputStream (isPOST, nullptr, nullptr,
+        auto urlStream = url.createInputStream (isPOST, nullptr,
                                                 accountEnquiryTask->getExtraHeaders(),
                                                 5000, nullptr, &statusCode);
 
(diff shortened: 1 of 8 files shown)


kunitoki:
Awesome, i was about to reopen it and you beat me to it.

Since it contains a breaking change in createInputStream i was thinking of making the original method JUCE_DEPRECATED as it will be easy to forward it to the std::function version.

kunitoki:
Opened the same on develop if it makes the process of merging faster.

https://github.com/juce-framework/JUCE/pull/831

JordanSkousen:
Awesome. I just tried adding back the old `createInputStream` method with `JUCE_DEPRECATED`, but there's JUCE code that only uses the first parameter of `createInputStream`, so the call becomes ambiguous.

kunitoki:
Yeah excessive use of default parameters. For functions with more than 4-5 like createInputStream i would have used a `struct CreateOptions`

shanekoss:
Can we get this merged?  Spent many hours chasing this down only to find it was a JUCE issue.  Thanks @JordanSkousen  and @kunitoki for this work - your fixes work great.

JordanSkousen:
+1 @shanekoss! Please, this is a bugfix! Not a suggestion or something!

JordanSkousen:
Changed the title so there's no doubt this is a bug 😉

ed95:
Closing this - see here https://github.com/juce-framework/JUCE/pull/831#issuecomment-814012113.

PR #831, opened 2021-01-15
Fix URL issue removing get parameters and adding them to post data (develop)

kunitoki:
When you have post data while using `createInputStream`, JUCE will end up removing the get parameters from the endpoint, and append them to the body, breaking both the endpoint to reach and the payload.

This also fix `createInputStream` to take a `std::function` for the progress callback instead of a function pointer and void* context.


kunitoki:
Sorry you removed juce6 branch and made me reopen the PR just to let it rot again for years?

ed95:
This is documented in the docs for `URL::createInputStream()` under the [`doPostLikeRequest` parameter](https://github.com/juce-framework/JUCE/blob/ba2027497b348a7fbe14f89cfde6666cb963983c/modules/juce_core/network/juce_URL.h#L315) so changing the behaviour as proposed would break existing code which may rely on this.

Perhaps a better solution would be to add methods for adding generic "body data" to a `juce::URL` similar to the existing `withPOSTData()` methods and then adding this data to the request body if present without also adding the parameters. Then users could call `createInputStream()` with `doPostLikeRequest=false` and set the `httpRequestCmd` to POST or PUT to send this data without the existing POST-like behaviour. Does this sound sensible? Would it work for your use-case?

kunitoki:
The behaviour of doPostLikeRequest is silly, no server implementation is going to parse querystring parameters as raw body data, ever (unless you want to parse the querystring manually extracting it from the body data and it's not a common use case). 

I would say i really like to be able to control what body data i send and the url should always respect the query string set in the URL class when doing the request (if you don't want the query string in the request, it's easy to create a new URL instance without it). Also i would remove the bool parameter, nobody will complain, and they could do with the new method.

ed95:
> Do you have a use case for this?
We use it in the Projucer licence query code - https://github.com/juce-framework/JUCE/blob/cc9fdc3d6a89b86ecd902c1f24258ae8cdcc5639/extras/Projucer/Source/Application/UserAccount/jucer_LicenseQueryThread.h#L325. It's a fair assumption that plenty of JUCE users are using it too, so we don't want to break their code.

kunitoki:
Yes but you can change that code to use your custom body data with encoded parameters placed before the real payload with the new withPOSTData method you are proposing to create.

To me the URL class has an API that makes too many assumptions and it's producing not always a what you write is what you get. Like not being able to make a POST request with a custom clean non tainted payload and without having it removing the query string is a big showstopper for its proper usage.

Please find a way to improve the API, don't complain user code might break if you are already pushing breaking changes for other things, we can fix our code if you are good enough to make it a non silent breaking change.

ed95:
We do push breaking changes, but they tend to be for code which is clearly incorrect or needs to be modernised to make it safer or clearer. In this case it's not clear to me that the API is "wrong" necessarily, it is just written in a particular way and, for better or worse, has been used by JUCE users for years and is deeply integrated into a lot of networking code written with JUCE.

I think we can keep the existing behaviour but also fit your use case with the proposed `withBodyData()` change that I suggested above, which is why I asked whether it would fit your use case. If it sounds sensible, then we could look at implementing something like that in the `URL` class as a way to preserve existing code but make the API more flexible.

kunitoki:
I'm all for it. Be sure we could use a untainted post body together with a untainted full (url with query string) when making a POST or PUT request

ed95:
I've pushed some changes to the `develop` branch which should address these issues: https://github.com/juce-framework/JUCE/commit/fa8c446d0cb3f6f956079afaa9e36cbb0f99b3ef

There's now a `URL::InputStreamOptions` class and a `URL::createInputStream()` overload which takes one of these as its argument and supersedes the old method with lots of parameters. The `ParameterHandling` enum allows you to specify whether the parameters will be transferred via the request body data, or via the URL address.

Now if you have a `URL` containing some POST data set via `URL::withPOSTData()` you can specify `ParameterHandling::inAddress` when calling `createInputStream()` and the request body will contain just the data, with the full URL containing the query string. This will allow you to send body data via POST or PUT requests without JUCE adding anything.

ed95:
I'm going to close this as I believe it has now been addressed with https://github.com/juce-framework/JUCE/commit/fa8c446d0cb3f6f956079afaa9e36cbb0f99b3ef, https://github.com/juce-framework/JUCE/commit/4ea3053b, and https://github.com/juce-framework/JUCE/commit/77d4ac72. Please re-open this if you feel further changes are needed.

PR #832, opened 2021-01-15
Better environment control when spawning processes

kunitoki:
Add support for better environment variables in the framework, because since today it was not possible to start a process and control which environment should have:

- Added functions to `SystemStats` to allow getting and changing the environment:
```
bool setEnvironmentVariable(const String& name, const String& value);

bool removeEnvironmentVariable(const String& name);

StringPairArray getEnvironmentVariables();
```

- Added functions to `ChildProcess` to pass the environment variables the child process should retain:
```
bool start (const String& command, const StringPairArray& environment, int streamFlags = wantStdOut | wantStdErr)
bool start (const StringArray& arguments, const StringPairArray& environment, int streamFlags = wantStdOut | wantStdErr)
```

- Added a function to `Process` to pass the environment variables the process spawned should retain:
```
static bool JUCE_CALLTYPE openDocument (
    const String& documentURL,
    const String& parameters,
    const StringPairArray& environment)
```

- Fixed macOS `Process::openDocument` to correctly spawn .app bundles with arguments and environment, and well as urls and local files.


modules/juce_core/native/juce_win32_Threads.cpp
```diff
@@ -445,15 +482,15 @@ class ChildProcess::ActiveProcess
 
             if (available == 0)
             {
-                if (! isRunning())
+                if (total != 0 || ! isRunning())
```
kunitoki, line 485: this is to produce the output faster on windows and avoid spinning if we alread have characters. this will make the shell output flow more fluently instead of getting out very slow and clunky on processes that produce lot of output (like build systems).

modules/juce_core/native/juce_win32_Threads.cpp
```diff
@@ -445,15 +482,15 @@ class ChildProcess::ActiveProcess
...
                     break;
-
+			
                 Thread::sleep (1);
             }
             else
             {
                 DWORD numRead = 0;
-                if (! ReadFile ((HANDLE) readPipe, dest, (DWORD) numToDo, &numRead, nullptr))
+                if (! ReadFile ((HANDLE) readPipe, dest, (DWORD) numToDo, &numRead, nullptr) || numRead == 0)
```
kunitoki, line 493: also this one

modules/juce_core/native/juce_win32_Files.cpp
```diff
@@ -951,6 +951,27 @@ bool JUCE_CALLTYPE Process::openDocument (const String& fileName, const String&
...
+        SystemStats::setEnvironmentVariable (key, environment.getValue (key, {}));
+    
+    HINSTANCE hInstance = ShellExecute (0, 0, fileName.toWideCharPointer(),
+                                        parameters.toWideCharPointer(), 0, SW_SHOWDEFAULT);
+
+    for (const auto& key : environment.getAllKeys())
+        SystemStats::removeEnvironmentVariable (key);
+
+    for (const auto& key : oldEnvironment.getAllKeys())
+        SystemStats::setEnvironmentVariable (key, oldEnvironment.getValue (key, {}));
```
kunitoki, line 970: unluckily ShellExecute inherit the environment and it doesn't allow to specify a custom environment like `execve`, so i had to modify the env in place

modules/juce_core/native/juce_posix_SharedCode.h
```diff
@@ -1143,24 +1218,22 @@ class ChildProcess::ActiveProcess
     {
         jassert (dest != nullptr && numBytes > 0);
 
-        #ifdef fdopen
-         #error // some crazy 3rd party headers (e.g. zlib) define this function as NULL!
-        #endif
-
-        if (readHandle == nullptr && childPID != 0)
-            readHandle = fdopen (pipeHandle, "r");
```
kunitoki, line 1151: by not doing this we can use use non buffering and non blocking `read` instead of `fread` which can block the process if no output is produced. speed is improved and you won't get character out in chunks with lot of latency and wait time between buffer are filled with new data. it is especially noticeable if you trigger cmake with lots of output from here.

kunitoki:
@tpoole thanks for pointing out the PR i made on juce6 has been closed automatically. Any chance these changes will be considered ?

PR #833, opened 2021-01-15
Enable github actions

kunitoki:
It would be nice if we could get these on GitHub, so we can keep track on what's missing for a PR to be fully releasable

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -22,8 +22,9 @@ jobs:
     - name: Build Debug
       shell: bash
       run: |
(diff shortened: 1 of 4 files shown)


tpoole:
Whilst your efforts to improve JUCE are certainly appreciated, we're not going to add GitHub actions to the repository.

Our internal CI suite consists of hundreds of jobs run on many different bits of hardware. It's not something we can open up to the public. Getting things to build on the relatively limited options that GitHub Actions can provide is nowhere near the coverage of platforms and build configurations that we require, and we don't want the overhead of maintaining a redundant CI pipeline.

PR #835, opened 2021-01-16
 Remove redundant condition

jerryuhoo:
I found a redundant condition "thisTokens.size() == 3"


ed95:
Thanks, fixed in https://github.com/juce-framework/JUCE/commit/1746d7bc4dae8d11eed78eb374060e075d7dad4e.

PR #839, opened 2021-01-18
Added a way to browse JUCE module sources in all projects

eyalamirmusic:
Even though the PR probably needs some adjustments and cleanup - this is (IMO) a much more comfortable way to browse sources in the IDE (XCode, Visual Studio, etc).

This is done without adding any additional targets, and only shows the module sources that are linked with a particular target. Works just as well with built in JUCE modules and custom modules.

Looking forward to your review!
Eyal
Screen Shot 2021-01-18 at 2 55 07 AM

### DIFF
--- a/extras/Build/CMake/JUCEReflectModuleSources.cmake
+++ b/extras/Build/CMake/JUCEReflectModuleSources.cmake
@@ -1,0 +1,42 @@
+#filters out the part that needs to compile from the module sources:
+function(_juce_clean_sources_list sourceList cleanList moduleName)
+    foreach (source IN LISTS sourceList)
+        get_filename_component(fileName ${source} NAME)
+
+        if (NOT fileName MATCHES ${moduleName}*)
+            list(APPEND localCleanList ${source})
+            set_source_files_properties(${source} PROPERTIES HEADER_FILE_ONLY TRUE)
+        endif ()
+    endforeach ()
+
+    set(${cleanList} "${localCleanList}" PARENT_SCOPE)
+endfunction()
+
+#A bit of a hack: Becuase CMake can't 'hold on' to the target properties here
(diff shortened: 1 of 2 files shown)


reuk:
I gave this a go, and adding this patch causes a significant performance regression when configuring the JUCE repo with `JUCE_BUILD_EXTRAS=ON` and `JUCE_BUILD_EXAMPLES=ON`.

```
$ rm -r cmake-build-xcode
$ git checkout origin/develop
$ time cmake -B cmake-build-xcode -G Xcode -D JUCE_BUILD_EXTRAS=ON -D JUCE_BUILD_EXAMPLES=ON
# output truncated
9.90s user 4.99s system 30% cpu 49.335 total

$ rm -r cmake-build-xcode
$ git checkout eyal/develop
$ time cmake -B cmake-build-xcode -G Xcode -D JUCE_BUILD_EXTRAS=ON -D JUCE_BUILD_EXAMPLES=ON
# output truncated
86.48s user 15.19s system 76% cpu 2:13.14 total
```

Reconfiguring an existing project takes about 12 seconds without this patch, and more than 90 seconds with this patch applied.

The feature does seem to work, which is good. After playing with it a bit, it looks like the slow part is the call to `target_sources` which adds the header files to the module target. Removing this single line brings the build times back to their original levels.

As it is, I think this approach is too slow to merge.

eyalamirmusic:
Updated the PR with a big performance improvement.
It's still slightly costly so possibly needs to be optional for JUCE/all targets, and definitely completely off for the JUCE build tools.

reuk:
Thanks for your work on this feature, added in https://github.com/juce-framework/JUCE/commit/7c88dec30d8a95e264dbf63ce6714c13d4e14c2c

PR #840, opened 2021-01-20
fix WebBrowserComponent file:// support for sandboxed apple apps

modosc:
this allows sandboxed apps on macos/ios to open `file://` urls in a `WebBrowserComponent` with `WkWebView`, see [here](https://stackoverflow.com/questions/24882834/wkwebview-not-loading-local-files-under-ios-8/26054170) for background.


ed95:
Thanks for the suggestion, this has been added for both the newer `WKWebView` and older `WebView` implementations in https://github.com/juce-framework/JUCE/commit/9622fc8e057b4654c06f77e4007ea55a73f50469.

PR #849, opened 2021-02-04
Fix juce_recommended_lto_flags broken on Visual Studio

danra:


extras/Build/CMake/JUCEHelperTargets.cmake
```diff
@@ -52,11 +52,11 @@ add_library(juce::juce_recommended_lto_flags ALIAS juce_recommended_lto_flags)
 if((CMAKE_CXX_COMPILER_ID STREQUAL "MSVC") OR (CMAKE_CXX_SIMULATE_ID STREQUAL "MSVC"))
     target_compile_options(juce_recommended_lto_flags INTERFACE
         $<$<CONFIG:Release>:$<IF:$<STREQUAL:"${CMAKE_CXX_COMPILER_ID}","MSVC">,-GL,-flto>>)
-    target_link_libraries(juce_recommended_lto_flags INTERFACE
+    target_link_options(juce_recommended_lto_flags INTERFACE
```
McMartin, line 55: `target_link_options` was introduced in CMake 3.13, but JUCE requires CMake 3.12, so this code cannot use `target_link_options` yet.

extras/Build/CMake/JUCEHelperTargets.cmake
```diff
@@ -51,9 +51,9 @@ add_library(juce::juce_recommended_lto_flags ALIAS juce_recommended_lto_flags)
 
 if((CMAKE_CXX_COMPILER_ID STREQUAL "MSVC") OR (CMAKE_CXX_SIMULATE_ID STREQUAL "MSVC"))
     target_compile_options(juce_recommended_lto_flags INTERFACE
-        $<$<CONFIG:Release>:$<IF:$<STREQUAL:"${CMAKE_CXX_COMPILER_ID}","MSVC">,-GL,-flto>>)
+        $<$<CONFIG:Release>:$<IF:$<STREQUAL:"${CMAKE_CXX_COMPILER_ID}","MSVC">,/GL,-flto>>)
```
McMartin, line 54: cl.exe supports both `-` and `/` for options. There is no "wrong form".

danra:
Ah, interesting.
Still, on VS (tested with 2019) the previous code just doesn't work, link time code generation is not enabled

danra:
Thanks, didn't know that.

Still would go with slash here for consistency (especially in the same file)

Also, since tons of other flags get added with slash, I think using slash for the flags you add yourself makes more sense, easier on the eyes when looking at the command line

McMartin:
The current code works fine. 

If I remove the `target_link_libraries` call completely, i.e. I don't pass `-LTCG` to the linker, it outputs:
```
  Main.obj : MSIL .netmodule or module compiled with /GL found; restarting link with /LTCG; add /LTCG to the link command line to improve linker performance
```
and still uses link time code generation.

If I remove the `target_compile_options` call (but keep the `target_link_libraries` call), i.e. I don't pass `-GL` to the compiler, the linker outputs:
```
  LINK : /LTCG specified but no code generation required; remove /LTCG from the link command line to improve linker performance
```
which proves that it was asked for link time code generation.

danra:
Ok, will double check. Maybe it did work and I thought it didn't just because VS didn't parse it as it did /LTCG.
Thanks!

danra:
@McMartin Indeed, LTCG does work in Visual Studio with current code. It's just uglier, because `"-LTCG"` appears under "Additional Dependencies" with other `.lib`s, instead of showing the "Link-time code generation" option as enabled.

Looks like CMake's `target_link_libraries` [documentation](https://cmake.org/cmake/help/latest/command/target_link_libraries.html) warns about this for the general case:

> Link flags specified here are inserted into the link command in the same place as the link libraries. This might not be correct, depending on the linker. Use the LINK_OPTIONS target property or target_link_options() command to add link flags explicitly. The flags will then be placed at the toolchain-defined flag position in the link command.

But I guess it's not relevant for JUCE at least with the current linkers it handles. Closing the PR.

PR #852, opened 2021-02-15
Change Mac goToURL behaviour to only call stop() if we load a new URL

tomduncalf:
Calling `stop` before evaluating Javascript will interrupt any assets loading on the current page, which is not the desired behaviour (it will result in e.g. images being half loaded, which if the Javascript does not result in loading a new page, can break the web content)


ed95:
https://github.com/juce-framework/JUCE/commit/0fc0dce1b8623d55e7e1d4bffec73d958936da45

PR #853, opened 2021-02-16
[Bugfix] Fix varArrayToNSArray and varObjectToNSDictionary

danra:
Fix `varArrayToNSArray` and `varObjectToNSDictionary` handling of nested arrays.

`var::isObject()` returns `true` for array variants, so it must be tested after `var::isArray()`.

### DIFF
--- a/modules/juce_core/native/juce_ObjCHelpers_mac.h
+++ b/modules/juce_core/native/juce_ObjCHelpers_mac.h
@@ -102,17 +102,17 @@ inline NSDictionary* varObjectToNSDictionary (const var& varToParse)
 
             const var& valueVar = properties.getValueAt (i);
 
-            if (valueVar.isObject())
-            {
-                auto* valueDictionary = varObjectToNSDictionary (valueVar);
-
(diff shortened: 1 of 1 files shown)


danra:
Ping. This is a straightforward bugfix. Please merge it in.

Anthony-Nicholls:
Hi @danra thanks for reporting. I've been looking into this. I've added some tests and I'm able to reproduce a bug likely caused by the issue you've raised. I think I can make some of the functions more resiliant than they currently are and probably remove a few of the functions too. However, I would like to know more about the issues you are facing with these functions. These functions are not meant to be part of the user facing library they are internal helpers. Are you using these functions directly? or did you encounter a bug via an indirect use of them? Thanks.

danra:
It's been a long time - I don't think I've used these directly, rather that I've seen that the end result of converting a var had some `NSObjects` internally instead of the `NSArray/Dictionary` objects I expected

Anthony-Nicholls:
Hi @danra, thanks for reporting. A fix for this has been merged into develop. Please let me know if you experience any more issues with this code.

PR #854, opened 2021-02-17
DSP: set Panner center value to 0 in documentation

tchissR:
Panner center value is set to 0.5 in documentation. Since the pan value is between -1 (left) and +1 (right), the center value is 0 and not 0.5.


ed95:
Thanks for reporting - added in https://github.com/juce-framework/JUCE/commit/51cda82fb75eb70db50946b7d28b45a5e5cb5908.

PR #857, opened 2021-02-18
Fix AUv3 multi-bus rendering bug

essej:
This fixes an issue where AUv3 plugins with multiple output buses were not rendering their additional output buses. The host calls the callback repeatedly for each output bus during a single process cycle, but the timestamp check was only allowing the first bus output to be rendered. With the fix, the first callback will cause everything to be computed, and the following callbacks the host makes with the same timestamp will just copy the already computed bus outputs that the host requested.

Also, the events should only be processed once at the first callback for this timestamp.

### DIFF
--- a/modules/juce_audio_plugin_client/AU/juce_AUv3_Wrapper.mm
+++ b/modules/juce_audio_plugin_client/AU/juce_AUv3_Wrapper.mm
@@ -1436,10 +1436,11 @@ private:
 
         // process params
         const int numParams = juceParameters.getNumParameters();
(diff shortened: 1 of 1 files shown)


reuk:
Thanks for raising this. I've made some changes here that should improve the situation: https://github.com/juce-framework/JUCE/commit/6de3af35665206383b47d48038ce0960dc9d509d

Please try updating and let us know if you encounter any further issues.

I think the issue should be resolve so I'm going to close this PR. I'm happy to reopen it (or move discussion to an Issue instead) if you run into any problems with the change.

PR #864, opened 2021-03-02
Check if controller is valid before destroying `boundsUpdater`

adamski:
In our iOS app, we are managing the `UIViewController` that holds a `JuceUIView`. Since updating to JUCE 6, we experienced crashes when calling `removeFromDesktop` on our JUCE components show in `UIView`s managed by our app. 

Checking that the `controller` is in fact a `JuceUIViewController` fixed the crash.


adamski:
Looks like this has been fixed in develop!

PR #865, opened 2021-03-03
Improve popup box keyboard control

miketoon:
Set ticked/selected item as highlighted on menu construction and
prevent fake mouse events from resetting this.

Previously when a menu was opened with the enter key a user would have to press escape key to close the menu. This was inconsistent if the selected/ticked item was highlighted using the arrow keys and then enter pressed.

Also previosuly keyboard navigation would start at the top/bottom of the list depending on whether up/down was pressed but this change uses the ticked/selected item as the start point.

### DIFF
--- a/modules/juce_gui_basics/menus/juce_PopupMenu.cpp
+++ b/modules/juce_gui_basics/menus/juce_PopupMenu.cpp
@@ -257,7 +257,12 @@ struct MenuWindow  : public Component
             auto& item = menu.items.getReference (i);
 
             if (i + 1 < menu.items.size() || ! item.isSeparator)
-                items.add (new ItemComponent (item, options, *this));
+            {
+                auto* child = items.add (new ItemComponent (item, options, *this));
+
+                if (item.itemID == options.getHighlightedItem())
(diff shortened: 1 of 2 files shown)


reuk:
I think this feature makes sense, but I'm not sure about the implementation. It's valid for a popup menu to have multiple ticked items, for example if it's created as a window menu, rather than a combobox dropdown. I don't think it makes sense to use the 'ticked' state to set the initial selection.

Another option might be to add an 'initially selected item' member to `PopupMenu::Options`, and to have the ComboBox automatically set this option to the currently selected item when showing the menu. This would also allow users to retain the existing menu behaviour, in the case that they don't wish to have any initial selection.

Do you think a solution like that would work for you?

miketoon:
yep that works. i've pushed a change to this branch which implements that.

reuk:
I've added this in https://github.com/juce-framework/JUCE/commit/d62d3aaa4f7803fe2c1175eb0ed6e65dc6077501.

Hopefully this works for you, please let me know if there are any issues.

PR #882, opened 2021-03-27
Add option to build modules as static/shared libs

anton-danielsson:
I'm working on a Conan package manager recipe for JUCE https://github.com/conan-io/conan-center-index/compare/master...DiracResearch:new/juce.
Still very much work in progress :)

One of the first things I ran into was that JUCE doesn't really build libraries.
I started implementing a work-around in the Conan recipe based on the same approach as:
https://forum.juce.com/t/using-cmake-flow-to-build-libraries-android-apps/40492/2.
But then I though that it would be much better if JUCE had an option to build libs,
others might benefit from this also.

This PR contains one approach to enable this. I would like to get some feedback if you think
this is a way forward or if you think this should be handled in the Conan recipe instead?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,6 +89,15 @@ if(JUCE_BUILD_EXAMPLES)
     add_subdirectory(examples)
 endif()
 
+# Normally JUCE creates modules as INTERFACE targets.
+# By turning this option on modules will instead be built as libraries.
+# CMake will create static or shared libs depending on BUILD_SHARED_LIBS.
+
+option(JUCE_BUILD_LIBRARIES
+    "Enable to build libraries (static or shared) of modules instead of source interface targets"
+    OFF
+)
+
 # ==================================================================================================
 # Install configuration
 
--- a/extras/Build/CMake/JUCEUtils.cmake
+++ b/extras/Build/CMake/JUCEUtils.cmake
@@ -180,8 +180,8 @@ _juce_set_default_properties()
 
 # ==================================================================================================
 
-function(_juce_add_standard_defs juce_target)
-    target_compile_definitions(${juce_target} INTERFACE
+function(_juce_add_standard_defs juce_target visibility)
+    target_compile_definitions(${juce_target} ${visibility}
         JUCE_GLOBAL_MODULE_SETTINGS_INCLUDED=1
         $<IF:$<CONFIG:DEBUG>,DEBUG=1 _DEBUG=1,NDEBUG=1 _NDEBUG=1>
         $<$<PLATFORM_ID:Android>:JUCE_ANDROID=1>)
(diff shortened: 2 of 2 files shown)


reuk:
I think this is unlikely to be practical for most users. JUCE has a large number of configuration flags, which are set using preprocessor definitions. Each project which depends on JUCE uses a slightly different collection of flags. Additionally, targets in the same build tree which both depend on JUCE may require a different set of flags. Building a static/shared lib forces all consumers of the library to use the same set of configuration flags, but it's impossible to pick a single 'canonical' set of flags.

Additionally, some JUCE modules use `JUCE_MODULE_AVAILABLE` preprocessor definitions to check for other modules which are being built into the same project. Such modules may reconfigure themselves depending on other modules which are present. This requires that all modules are built with the correct set of `MODULE_AVAILABLE` flags, which isn't possible if each module is built into a separate library target.

anton-danielsson:
Maybe not for most but for some it might be useful.
I'm a long time JUCE user and love it. But I think the build system is a bit quirky (and that's still true in JUCE 6 and CMake)
To give some background I can share with you that we made our own CMake scripts for JUCE more then 7 years ago and we have used our own Conan recipe for a few years now. So I'm quite familiar with how to build JUCE :)
And I can say that a library based approach fits our needs better.

When it comes to practicalities like flags and options I think Conan is perfectly capable of handling this.
Conan already have recipes for projects like Qt and Boost that provides similar challenges :)
Which modules are included and the JUCE flags will be handled with recipe options.
This includes things like `JUCE_ASIO`  and `JUCE_USE_CURL`.
And since modules will be part of the recipe options the `JUCE_MODULE_AVAILABLE` will not be a problem
as all modules will be built with the same set of `MODULE_AVAILABLE` flags.

The demand for a JUCE Conan package is maybe not huge but I have seen some requests like:
https://forum.juce.com/t/juce-conan-package/27640
https://github.com/conan-io/wishlist/issues/59
There are also a couple of recipes out there but they appear un-maintained:
https://github.com/khdkhd/conan-juce
https://github.com/pixsperdavid/conan-juce

My goal is to get the recipe into Conan Center Index and make sure it stays updated.
One thing that will help keeping it updated is to minimize the number of patches and that
is why I want to discuss a few changes here that might make the Conan recipe easier to write.
At the moment I'm trying a few approaches and it might prove that the way described in
https://forum.juce.com/t/using-cmake-flow-to-build-libraries-android-apps/40492/2
will work better.
Building one DLL per module will for example be a challenge on Windows due to
how JUCE_API and JUCE_DLL_BUILD/JUCE_DLL is setup... Making that work feels like major pain...
So it's probably better to just build one DLL of JUCE, even though Conan like to have the
recipes modularized using "components".

PR #885, opened 2021-04-02
Paste from the correct X11 selection

robbert-vdh:
Pasting text on Linux in JUCE plugins and applications has never worked correctly. Pasting would always be done from the `PRIMARY` selection instead of the `CLIPBOARD` selection, even if the comment right this did say that pasting should always be done from the `CLIPBOARD` selection like one would expect. This would cause issues when trying to paste passwords from a password manager, or when trying to paste serial numbers from websites copied using a copy button (i.e. without selecting the entire serial number first).

https://forum.juce.com/t/juce-uses-the-wrong-x11-selection-for-pasting/45330


ed95:
Thank you for reporting and for the proposed fix. We've fixed this on the `develop` branch in https://github.com/juce-framework/JUCE/commit/faf5ed40234c038c67c15220400f3ffacd4b323c.

PR #891, opened 2021-04-19
add ladspa-sdk to Linux dependencies

mynameismuhl:
cmake errors otherwise


ed95:
Added in https://github.com/juce-framework/JUCE/commit/b7647e2acb0f89265ec818addc73eb01b4f9f054. Thanks

PR #896, opened 2021-04-29
Adding 0 elements to an empty array resulted in undefined behaviour

omartijn:
If the array is empty and you're pushing zero elements, no pointer is
allocated (because it's not needed), but then a nullptr is given to
memcpy, which is explicitly defined as UB in the standard.


reuk:
Thanks, added in https://github.com/juce-framework/JUCE/commit/22189f7aa435c38f68f739b84566a847acf625d0

PR #904, opened 2021-05-31
X11: Fix DPI scaling for KDE systems

anthroid:
KDE systems can have both `/usr/bin/dconf` and/or `/usr/bin/gsettings` installed as dependencies for GTK apps, and these can (and likely will) return invalid scale factors, resulting in tiny UI scaling on HiDPI systems. This adds support for KDE by querying the global scale factor from `/bin/kreadconfig5`, which reflects the value in System Settings, instead of a likely erroneous value from `dconf` or `gsettings`. Due to the possibility and likelihood of all 3 existing on a KDE system, this checks for the existence of `/bin/kreadconfig5` before checking for `dconf` or `gsettings`.

### DIFF
--- a/modules/juce_gui_basics/native/x11/juce_linux_XWindowSystem.cpp
+++ b/modules/juce_gui_basics/native/x11/juce_linux_XWindowSystem.cpp
@@ -1102,15 +1102,41 @@ namespace DisplayHelpers
 
         if (scaleFactorFromXSettings > 0.0)
             return scaleFactorFromXSettings;
-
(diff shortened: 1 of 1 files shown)


anthroid:
Closing for now. This fixes scaling in the Projucer and the example apps, but VST3 plugins (and potentially others) will open initially with the correct scale factor, but then subsequent opens in the same session (reopening the GUI without deleting and reinstantiating the plugin) do not use the correct DPI (tested in Bitwig and Renoise on Arch + KDE).

PR #911, opened 2021-06-16
juce::Colour HSL support: Skip unnecessary calculations, avoid division by zero

marcusfreisleben:
Fixes the hue/saturation conversion code to skip further calculations if `min(r,g,b) == max(r,g,b)`, since hue (and saturation) is always zero in that case. This also avoids a division by zero due to `invDiff = hi - lo = 0` for e.g. plain white.


reuk:
Thanks for reporting, I've fixed these issues here: https://github.com/juce-framework/JUCE/commit/f0c02529ef5fcc4a2c5aedc27f6366afe9fa2215

PR #920, opened 2021-07-16
Prevent unused variable warning using jassert in release

omartijn:
When a variable is declared and then only used inside a `jassert`
statement, the compiler rightfully complains about the variable being
unused when building in release mode - since in that case `jassert`
doesn't do anything.

This change will "use" the variable even in that case, but careful use
of constexpr and short-circuiting ensures the expression is never
actually executed, which is important to avoid both performance issues
when the thing being asserted is a heavy expression (which you only do
in debug) or when the expression itself has side-effects.

See https://forum.juce.com/t/make-jassert-use-ignoreunused-internally/36728 for discussion

### DIFF
--- a/modules/juce_core/system/juce_PlatformDefs.h
+++ b/modules/juce_core/system/juce_PlatformDefs.h
@@ -171,7 +171,7 @@ namespace juce
   #if JUCE_LOG_ASSERTIONS
    #define jassert(expression)      JUCE_BLOCK_WITH_FORCED_SEMICOLON (if (! (expression)) jassertfalse;)
   #else
-   #define jassert(expression)      JUCE_BLOCK_WITH_FORCED_SEMICOLON ( ; )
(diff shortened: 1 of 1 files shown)


modules/juce_core/system/juce_PlatformDefs.h
```diff
@@ -171,7 +171,7 @@ namespace juce
   #if JUCE_LOG_ASSERTIONS
    #define jassert(expression)      JUCE_BLOCK_WITH_FORCED_SEMICOLON (if (! (expression)) jassertfalse;)
   #else
-   #define jassert(expression)      JUCE_BLOCK_WITH_FORCED_SEMICOLON ( ; )
+   #define jassert(expression)      JUCE_BLOCK_WITH_FORCED_SEMICOLON ( [[maybe_unused]] constexpr bool jassert_unused = false && (expression); )
```
yairchu, line 174: Is there a reason to prefer `[[maybe_unused]]` over `juce::ignoreUnused` (which has wider C++ support)?

PR #922, opened 2021-07-21
GUI not shown in Mac Audition & Premiere

mootensai:
The problem not only exists in Audition but in Premiere as well.


ed95:
Thanks for reporting. We've added this in https://github.com/juce-framework/JUCE/commit/4fd0401752f53d418b5108f220cbac685cafdcfd.

PR #931, opened 2021-08-05
Fixes Linux x11 compilation bug with missed unique_ptr get()'s

bboettcher3:


ed95:
This is fixed on the `develop` branch.

PR #932, opened 2021-08-12
Add missing VST3 poly aftertouch support

falkTX:
Initially reported at https://github.com/falkTX/Carla/issues/1386
Issue (missing aftertouch) confirmed fixed after this change.

### DIFF
--- a/modules/juce_audio_processors/format_types/juce_VST3Common.h
+++ b/modules/juce_audio_processors/format_types/juce_VST3Common.h
@@ -957,6 +957,12 @@ private:
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


falkTX:
I did not reproduce this myself, as mentioned in the ticket.
(Do not have such hardware to quickly test this, and did not setup tool to generate it in software)

My understanding of the VST3 spec is limited, so I am not quite sure this is correct. Just passing some information along.

mkruselj:
This PR can be closed. Poly aftertouch messages already work with JUCE generated VST3s - Surge XT is a fine example showing it working!

PR #934, opened 2021-08-16
Adding a way to draw outer- and inner-shadows of paths using the StackBlur algorithm

LukeSonnox:


modules/juce_graphics/effects/juce_StackShadowEffect.cpp
```diff
@@ -0,0 +1,329 @@
...
+// mario@quasimondo.com and described here:
+// http://incubator.quasimondo.com/processing/fast_blur_deluxe.php
+
+// Stackblur algorithm by Mario Klingemann
+// Details here:
+// http://www.quasimondo.com/StackBlurForCanvas/StackBlurDemo.html
+// C++ implemenation base from:
+// https://gist.github.com/benjamin9999/3809142
+// http://www.antigrain.com/__code/include/agg_blur.h.html
+static void applyStackBlur (Image& img, unsigned int radius)
```
ImJimmi, line 83: To be in accordance with other JUCE APIs, could this just be `int radius`, and an assertion added to check that it's >= 0.

modules/juce_graphics/effects/juce_StackShadowEffect.cpp
```diff
@@ -0,0 +1,329 @@
...
+    unsigned long sum, sum_in, sum_out;
+
+    unsigned int wm = w - 1;
+    unsigned int hm = h - 1;
+    unsigned int w1 = (unsigned int) data.lineStride;
+    unsigned int div = (unsigned int) (radius * 2) + 1;
+    unsigned int mul_sum = stackblur_mul[radius];
+    unsigned char shr_sum = stackblur_shr[radius];
+
+    for (y = 0; y < h; ++y)
```
ImJimmi, line 109: These loop variables should be declared in the loop itself, not outside it.

```
for (auto y = 0; y < h; y++)
```

modules/juce_graphics/effects/juce_StackShadowEffect.cpp
```diff
@@ -0,0 +1,329 @@
...
+        xp = radius;
+        if (xp > wm)
+            xp = wm;
+
+        src_ptr = data.getLinePointer (int (y)) + (unsigned int)data.pixelStride * xp;
+        dst_ptr = data.getLinePointer (int (y));
+
+        for (x = 0; x < w; ++x)
+        {
+            dst_ptr[0] = (unsigned char)((sum * mul_sum) >> shr_sum);
```
ImJimmi, line 144: Could reduce the size of the nested loops, and aid readability by encapsulating them in free functions, something like:

```
for (auto y = 0; y < h; y++)
{
    // ...

    for (auto x = 0; x < w; x++)
    {
        blurHorizontally (x, y);
    }
}
```

PR #938, opened 2021-08-24
Add default return value for Component::exitModalState to prevent the need to pass an unused argument.

ImJimmi:
When using `exitModalState()` after having called `enterModalState()` on a component, the argument passed to `exitModalState()` is unused. By providing a default return value, we're left with a neat `enter`/`exit` pair of methods.

```cpp
comp.enterModalState();
// ...
comp.exitModalState (0); // Unused magic number - bad!
```

```cpp
comp.enterModalState();
// ...
comp.exitModalState(); // No magic numbers - good!
```

### DIFF
--- a/modules/juce_gui_basics/components/juce_Component.h
+++ b/modules/juce_gui_basics/components/juce_Component.h
@@ -2081,7 +2081,7 @@ public:
 
         @see runModalLoop, enterModalState, isCurrentlyModal
     */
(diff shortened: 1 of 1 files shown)


tpoole:
Added in https://github.com/juce-framework/JUCE/commit/098262aa8af26c64b372dcd5de8658122d3794ef.

Thank you.

PR #949, opened 2021-09-12
Fix: Should not update the lastNonFullscreenBounds variable when ComponentPeer is in minimised.

COx2:
Hello JUCE team,

I found a strange behaviour when `juce::DocumentWindow` is using native title bar.
This PR fixes the strange behavior of maximizing and minimizing window operations.

How to reproduce.
+ Operating System: Windows 10
+ Use native title bar on document window by `juce::DocumentWindow::setUsingNativeTitleBar (true)`
+ Make window maximised by clicking maximise button. Window will be fullscreen.
+ Make window minimised by clicking Windows taskbar app icon. Window will be minimised.
+ Show window by clicking Windows taskbar app icon. Window will be fullscreen.
+ If you click the maximize button of the window, it will stay at the maximized size instead of returning to its original size.

I found a code below that can solve a similar problem.
https://github.com/juce-framework/JUCE/blob/master/modules/juce_gui_basics/windows/juce_ResizableWindow.cpp#L509

However, that code will not pass when using the native title bar, so I adopt a similar implementation in this PR.

Before applying this PR, the window will not return to the size it was before going fullscreen.
After applying this PR, the window will return to the size it was before going fullscreen.

Please check it.

### DIFF
--- a/modules/juce_gui_basics/windows/juce_ComponentPeer.cpp
+++ b/modules/juce_gui_basics/windows/juce_ComponentPeer.cpp
@@ -322,7 +322,7 @@ void ComponentPeer::handleMovedOrResized()
         component.sendVisibilityChangeMessage();
     }
 
-    if (! isFullScreen())
+    if (!(isFullScreen() || isMinimised() || isKioskMode()))
         lastNonFullscreenBounds = component.getBounds();
 }
 


COx2:
This video shows the behavior before applying the fix by this PR.

https://user-images.githubusercontent.com/2960956/132973603-0ba556b6-6afc-4fcb-8925-7de792d0f3b0.mp4

COx2:
This video shows the behavior after applying the fix by this PR.

https://user-images.githubusercontent.com/2960956/132973630-201acae7-00c2-42b4-91fc-13f67458355d.mp4

reuk:
Thanks for the detailed explanation and patch. I've added a fix for this issue here: https://github.com/juce-framework/JUCE/commit/9360c3f4fffcb107ff2ab3d1b09949a2b6b35f07

COx2:
Thank you for your code review.
Nice.

PR #955, opened 2021-10-06
Use onlyOld as option for CoreMIDI if not on right version

akivab:
Finding a crash on startup on iOS13 due to what I imagine is due __IPHONE_14_0 defined at compile time, but __IPHONE_OS_VErSION_MIN_REQUIRED == __IPHONE_13_0

### DIFF
--- a/modules/juce_audio_devices/native/juce_mac_CoreMidi.mm
+++ b/modules/juce_audio_devices/native/juce_mac_CoreMidi.mm
@@ -52,16 +52,10 @@ namespace CoreMidiHelpers
         onlyOld
     };
 
-    #if (defined (MAC_OS_VERSION_11_0) || defined (__IPHONE_14_0))
-     #if (MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_VERSION_11_0 || __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_14_0)
-      #define JUCE_HAS_NEW_COREMIDI_API 1
-      #define JUCE_HAS_OLD_COREMIDI_API 0
-      constexpr auto implementationStrategy = ImplementationStrategy::onlyNew;
-     #else
-      #define JUCE_HAS_NEW_COREMIDI_API 1
-      #define JUCE_HAS_OLD_COREMIDI_API 1
-      constexpr auto implementationStrategy = ImplementationStrategy::both;
-     #endif
+    #if (MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_VERSION_11_0 || __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_14_0)
+     #define JUCE_HAS_NEW_COREMIDI_API 1
+     #define JUCE_HAS_OLD_COREMIDI_API 0
+     constexpr auto implementationStrategy = ImplementationStrategy::onlyNew;
     #else
      #define JUCE_HAS_NEW_COREMIDI_API 0
      #define JUCE_HAS_OLD_COREMIDI_API 1


reuk:
I'm not able to reproduce the issue here. I don't have a physical device running iOS 13, so I'm testing with a 13.4 simulator.

To test, I configured a CMake build passing `-G Xcode -D CMAKE_OSX_DEPLOYMENT_TARGET=10.0`. I built with Xcode 12.5, which has the 14.4 SDK.

Then, I built and ran the MIDIDemo app on the 13.4 simulator. The app opens cleanly and doesn't crash.

Please can you provide some more details about how you're building your project. What deployment target, Xcode version, and iOS SDK are you using? Are you testing on a device or a simulator?

akivab:
The crash is happening on physical iOS devices due to an issue with dynamically linking the CoreMIDI library (to resolve kMIDIPropertyProtocolID) at startup.

Deployment target is 13.0, it builds with CMake and can be distributed so no issue there, seems to be some issue when actually running on a device (kMIDIPropertyProtocolID not found in the dynamic lib)

Incident Identifier: 1FEEFABD-9102-418F-B17D-039FF9C555B1
CrashReporter Key:   8ee3348004dffed7f0d0c0fd38ae226af9c38106
Hardware Model:      iPhone8,4
Process:             app [297]
Path:                /private/var/containers/Bundle/Application/B6726A9B-7D1F-4E56-9205-655010AB46FD/app
Identifier:          com.acme.app
Version:             482 (1.0.13)
AppStoreTools:       13A227
AppVariant:          1:iPhone8,4:13
Code Type:           ARM-64 (Native)
Role:                Foreground
Parent Process:      launchd [1]
Coalition:           app [446]
Date/Time:           2021-10-05 20:09:46.8707 -0700
Launch Time:         2021-10-05 20:09:46.6224 -0700
OS Version:          iPhone OS 13.6.1 (17G80)
Release Type:        User
Baseband Version:    9.71.01
Report Version:      104
Exception Type:  EXC_CRASH (SIGABRT)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Exception Note:  EXC_CORPSE_NOTIFY
Termination Description: DYLD, symbol ‘_kMIDIPropertyProtocolID’ not found, expected in ‘/System/Library/Frameworks/CoreMIDI.framework/CoreMIDI’, needed by ‘/private/var/containers/Bundle/Application/B6726A9B-7D1F-4E56-9205-655010AB46FD/app'
Highlighted by Thread:  0
Backtrace not available
Unknown thread crashed with ARM Thread State (64-bit):
    x0: 0x0000000000000006   x1: 0x0000000000000004   x2: 0x000000016f5a9440   x3: 0x00000000000000c8
    x4: 0x000000016f5a9040   x5: 0x0000000000000000   x6: 0x0000000000000000   x7: 0x0000000000000000
    x8: 0x0000000000000020   x9: 0x0000000000000009  x10: 0x70612e7469206b79  x11: 0x69206b79616d2f70
   x12: 0x3634424130313035  x13: 0x206b79616d2f4446  x14: 0x6d2f7070612e7469  x15: 0x00277469206b7961
   x16: 0x0000000000000209  x17: 0x00000001046c2eb8  x18: 0x0000000000000000  x19: 0x0000000000000000
   x20: 0x000000016f5a9040  x21: 0x00000000000000c8  x22: 0x000000016f5a9440  x23: 0x0000000000000004
   x24: 0x0000000000000006  x25: 0x00000001046c8000  x26: 0x0000000000000000  x27: 0x00000001046c8000

akivab:
I think this issue is caused by compiling for iOS where iOS 14 is defined but is not the min supported version

reuk:
Have you tried running any of the JUCE demo projects on a physical device, after building it with the same deployment target and SDK? If so, do the JUCE projects exhibit the same behaviour? I'm trying to establish whether this is an issue with all JUCE apps, or just your particular app.

The fix you suggested is not particularly viable - the "old" MIDI API has a tendency to crash on iOS 14.2, so we definitely want to always use the "new" API if it is available. In particular, if the deployment target is lower than 14.0, we still want to use the new MIDI API on devices running iOS 14.0 or higher. There are more details about this issue in [this thread](https://forum.juce.com/t/ios-14-2-midi-connections-broken/42440).

It looks to me like `kMIDIPropertyProtocolID` should be weak-linked when building with a deployment target lower than iOS 14.0. That certainly seems to be the case on macOS: If I build an app like this on macOS 11.0, then run it on macOS 10.15, the app exists successfully, printing "no symbol". It doesn't crash, despite the 10.15 SDK having no symbol corresponding to `kMIDIPropertyProtocolID`.

```cpp
#include <juce_core/juce_core.h>
#include <CoreMidi/CoreMidi.h>

int main() {
  if (&kMIDIPropertyProtocolID != nullptr)
    DBG("symbol found");
  else
    DBG("no symbol");

  return 0;
}
```

If your app is crashing during startup, then I think this might indicate that the symbol has incorrectly been strongly-linked. Have you tried testing with different versions of Xcode, and if so, do you see the same results with every version?

akivab:
Thanks @reuk, will close this for now as I investigate more.

PR #973, opened 2021-11-24
Set bypass flag for plugin

pixout:
Quite usable to have the possibility to set the DAW bypass flag for the plugin.


pixout:
I will try to pull into develop

PR #974, opened 2021-11-24
Set the DAW bypass flag for the plugin.

pixout:
Quite usable to have the possibility to set the DAW bypass flag for the plugin.

### DIFF
--- a/modules/juce_audio_processors/processors/juce_AudioProcessor.h
+++ b/modules/juce_audio_processors/processors/juce_AudioProcessor.h
@@ -796,6 +796,12 @@ public:
     */
     double getSampleRate() const noexcept                       { return currentSampleRate; }
 
+    /** Returns true if plugin in the current DAW is bypassed. */
+    bool getIsDawBypass() const noexcept                        { return isDawBypass; }
+
+    /** Changes the processor flag to bypassed. */
+    void setIsDawBypass(bool bIsDawBypass) noexcept             { isDawBypass = bIsDawBypass; }
+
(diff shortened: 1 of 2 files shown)


reuk:
Please can you provide some more details about what problem this is intended to solve?

pixout:
From time to time plugin needs to go in bypass mode when (for example) DAW is rendering and in this case, I propose to introduce new two functions because, in my opinion, this is functionality is more close to AudioProcessor than to config.

talaviram:
Am I'm missing something? Your commit only adds a member with getter/setter.

This can be achieved by your inherited AudioProcessor.

Have you looked into :

processBlockBypassed and the bypass parameter? Did you take into consideration plug-ins with delay lines needing to compensate when bypassed?

pixout:
Ok, maybe you are right and this code should be placed in inherited AutioProcessor, but anyway, having a bypass option will be very useful.

reuk:
This already exists via `getBypassParameter`. If you want to the host to notify the plugin about bypass state changes, override `getBypassParameter` to return some parameter. The plugin can then check this parameter's value to see whether or not it should be processing.

If you're trying to set the bypass state on a hosted plugin, you need to do the following:
- If the plugin's getBypassParameter returns non-null, set the bypass state by updating the parameter value.
- Otherwise, the host should call `processBlockBypassed` if the plugin is bypassed, or `processBlock` if the plugin is not bypassed.

I'd recommend carefully reading the documentation for all of the functions I just mentioned to get a better idea of how all the parts are supposed to fit together.

PR #996, opened 2022-01-13
Allow specifying a parent component for ContentShrarer

lucijan:
AudioUnits on iOS don't support components added to desktop. This patch will
add the popover to the specified parent component's peer instead.


lucijan:
This has been taken care of with commit 9d1a6a3b28f90c055905114d63a2edec932f472f

PR #1038, opened 2022-03-17
[Android] replace system ClassLoader with fully working ClassLoader.

atsushieno:
This fixes one of the issues exposed as https://forum.juce.com/t/env-findclass-does-not-work-inside-juce-app-very-strange/46886 .
A fully detailed explanation is available at https://atsushieno.github.io/2022/03/16/juce-android-threads.html

In the current JUCE JNI invocation foundation (JNIClassBase etc.), the
Android system ClassLoader was used. It turned out that it is different from
the actual ClassLoader that we see in Dalvik bytecode VM, and the current
one fails to load non-system and non-pre-registered Java classes.
It can be reproduced by adding some code before call to `Java.initialiseJUCE
()`
in `JuceApp`:

```
	this.getClass().getClassLoader().loadClass("com.rmsl.juce.Java");
	ClassLoader.getSystemClassLoader().loadClass("com.rmsl.juce.Java");
	com.rmsl.juce.Java.initialiseJUCE(context.getApplicationContext());
```

A fully working ClassLoader can be retrieved from `com.rmsl.juce.Java` Class
using `getClassLoader()`. Use it instead of system ClassLoader.

This needs additional change on where to call `initialiseAllClasses()`
(move from `JNI_OnLoad()` to `Thread::initialiseJUCE()` as we need a
Dalvik bytecode VM.


atsushieno:
Oops my develop branch was not at HEAD. Let me rebase it.

PR #1039, opened 2022-03-17
[Android] invoke juce::Thread runnable from within java.lang.Thread context.

atsushieno:
This fixes the core part of the issues exposed at https://forum.juce.com/t/env-findclass-does-not-work-inside-juce-app-very-strange/46886
The full details are explained at https://atsushieno.github.io/2022/03/16/juce-android-threads.html

The existing juce::Thread invokes its runnable (`run()`) target using
`pthread_create()`. This is causing any JNI invocation that involves
non-framework classes and non-registered JUCE Java classes not loading.

This Android NDK documentation explains that it is anticipated:
https://developer.android.com/training/articles/perf-jni#threads

To fix this issue, `java.lang.Thread` context is needed.

Since there seems no way to first create `pthread_create()` and attach it to
`java.lang.Thread` without already having it running, we need to have a
`java.lang.Thread` and let it callback to native code (we can use `Runnable`
proxy based on JNIClassBase), retrieve `pthread_t` ID via `pthread_self()`
and proceeed to the actual runnable (`run()` on `juce::Thread`).

The JNI part is involved only at `start()` in this change - on Android
thread cancellation is not really supported (not in JUCE on Android either).


atsushieno:
same as #1038 let me rebase it (was not at develop HEAD).

PR #1048, opened 2022-04-01
[Android] Add missing activity lifecycle callback "onActivityConfigurationChanged"

ChrisBews:
We noticed this issue when upgrading our Pixel devices to the latest (March 2022) security update. This method was apparently not being called previously, but with the latest security update, it now it does, at least when the device is rotated. This now results in such devices experiencing a crash with the error:

`Expected receiver of type android.app.Application$ActivityLifecycleCallbacks, but got java.lang.Object`

Similar to previously mergd pr: https://github.com/juce-framework/JUCE/pull/574

Disclaimer: I am not a C++ dev, feel free to correct me :)


reuk:
I can reproduce the issue, and I've verified the fix. Weirdly, I can't find any information about the `onActivityConfigurationChanged` function, and it doesn't seem to be in the Android 31 `ActivityLifecycleCallbacks` interface, which worries me a bit. Hopefully there aren't other missing undocumented functions...

reuk:
Thanks for your patience, this is now merged to develop: https://github.com/juce-framework/JUCE/commit/d182ed38cd3c960a25034f642585063d91002a6f

ChrisBews:
Thanks for getting this change in Reuben! Worth the wait. I'll close this PR now.

PR #1062, opened 2022-04-26
handle canSelectDirectories in iOS native FileChooser

faqteur:
Use canSelectDirectories flag in iOS native FileChooser

### DIFF
--- a/modules/juce_gui_basics/native/juce_ios_FileChooser.mm
+++ b/modules/juce_gui_basics/native/juce_ios_FileChooser.mm
@@ -47,7 +47,7 @@ public:
         auto* controllerClassInstance = controllerClass.createInstance();
 
         String firstFileExtension;
-        auto utTypeArray = createNSArrayFromStringArray (getUTTypesForWildcards (owner.filters, firstFileExtension));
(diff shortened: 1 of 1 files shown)


izzyreal:
My initial try-out with this does not give the expected results. The file browser view changes from Icons to List (that I can't change to Icons view anymore) and the files are not displayed; only directories.

izzyreal:
Been playing a bit in an Objective C playground app and with `UIDocumentBrowserViewController` I get the desired result -- I can select multiple files and directories using this code:
```objc
#import "ViewController.h"

#include <CoreServices/UTType.h>
#import <UniformTypeIdentifiers/UniformTypeIdentifiers.h>

@interface ViewController ()
@end

@implementation ViewController

- (void)viewDidAppear:(BOOL)animated {
  NSArray *utis = @[ (NSString *)UTTypeItem ];
  UIDocumentBrowserViewController *picker =
    [[UIDocumentBrowserViewController alloc] initForOpeningContentTypes:utis];
  [self presentViewController:picker animated:YES completion:nil];
}

@end
```

The same code with `UIDocumentPickerViewController` reproduces JUCE's behaviour.

PR #1067, opened 2022-05-06
Always set sample rate of hosted AU plugins, fixes a bug with plugins…

seb-nektar:
The **AU** plugins from the **Roland Cloud** collection (e.g. [JX3P](https://www.roland.com/us/products/rc_jx-3p/)) report that they are already set to a sample rate of 44.1 kHz when they are first instantiated. This is incorrect and internally they are set to another default sample rate.

On the JUCE side, the AU hosting code compares the sample rate set in the hosted plugin with the new sample rate to be set. When both are the same, the sample rate is not set again and has the effect that Roland Cloud AU plugins play back at incorrect pitch when being instantiated in a JUCE host with 44.1 kHz sample rate.

While this clearly is a bug in the Roalnd plugins I think JUCE could handle that case more gracefully by simply always setting the sample rate during `prepareToPlay()` as shown in this pull request.


seb-nektar:
Duplicate of PR #1097

DUNYAM01:
Mmkkk

PR #1070, opened 2022-05-16
Small changes to handle large, multi-message sysEx data transfers

lucianthorr:
Hi. We are currently working on a project to use JUCE to handle firmware updates which require us to send large sysEx payloads that span multiple messages. In the process, we found that we needed to make a couple small changes to the JUCE library.

One change to the MidiMessage validation: We needed to allow small messages that could be the trailing end of a sysEx payload to pass the `jassert`

And secondly, we needed a way to force CoreMidi to use the old implementation because the new implementation does not currently handle large, multi-message sysEx payloads.

Thanks!
(resubmitting because the CLA verification failed on the first try)


lucianthorr:
I signed and submitted the CLA before making the PR.  Maybe there's an problem because my work email is not my primary Github email address?

PR #1090, opened 2022-07-11
Remove some unnecessary std::initializer_list temporaries.

omartijn:
We don't need them, since `ignoreUnused` will happily take as many
parameters as we give it. We don't need the `ignoreUnused` at all for
functions returning void - there is nothing to ignore, after all.

Thank you for submitting a pull request.

### DIFF
--- a/modules/juce_core/containers/juce_ArrayBase.h
+++ b/modules/juce_core/containers/juce_ArrayBase.h
@@ -557,7 +557,7 @@ private:
     template <typename... Elements>
     void addImpl (Elements&&... toAdd)
     {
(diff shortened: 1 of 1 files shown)


reuk:
In a function call, value computations and side effects of the initialization of every parameter are indeterminately sequenced with respect to value computations and side effects of any other parameter.

In [list-initialization](https://en.cppreference.com/w/cpp/language/list_initialization), every value computation and side effect of a given initializer clause is sequenced before every value computation and side effect associated with any initializer clause that follows it in the brace-enclosed comma-separated list of initalizers.

I'm pretty sure we want to use list initialization here (at least for `addAssumingCapacityIsReady`) to ensure that the arguments are added in the expected order.

PR #1101, opened 2022-08-09
Added an overload to sort a ValueTree using a lambda

ffAudio:
This Pull Request adds an function overload to ValueTree sort, which takes a lambda, so that sorting of ValueTree children can be done in one line instead of creating an ElementComparator struct.

Unit Tests for numbers and strings are included.

### DIFF
--- a/modules/juce_data_structures/values/juce_ValueTree.cpp
+++ b/modules/juce_data_structures/values/juce_ValueTree.cpp
@@ -1094,6 +1094,23 @@ ValueTree ValueTree::readFromGZIPData (const void* data, size_t numBytes)
     return readFromStream (gzipStream);
 }
 
+struct LambdaSorter
+{
+    LambdaSorter (ValueTree::SortLambda s) : sortFunc (s) {}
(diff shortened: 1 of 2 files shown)


reFX-Mike:
Very nice. Perhaps something similar can be done for StringArray, ListBoxModel, etc., wherever a custom sort might be needed.

ffAudio:
> Very nice. Perhaps something similar can be done for StringArray, ListBoxModel, etc., wherever a custom sort might be needed.

Indeed, this would be possible. Good idea. `juce::StringArray` just sits on top of `juce::Array<juce::String>` and Array has a sort using a similar ElementComparator. So the templated variant I had before would be doable for juce::Array.
The ListBoxModel is abstract though, it is not really a container, so I think here it doesn't really apply.

ffAudio:
Bump with backlink to the forum
https://forum.juce.com/t/std-sort-on-juce-valuetree/52532/10

ffAudio:
Anybody at juce?

ffAudio:
Thanks for the feedback

PR #1110, opened 2022-08-25
Extended the RuntimePermissions API to support any string.

jrlanglois:


modules/juce_core/misc/juce_RuntimePermissions.h
```diff
@@ -113,11 +113,13 @@ class JUCE_API  RuntimePermissions
         the user has responded).
     */
     static void request (PermissionID permission, Callback callback);
+    static void request (const juce::String& permission, Callback callback);
```
jrlanglois, line 116: I've left the docs out not knowing if copy/paste was the best option or not (minus some tweaks to document expectations about the string param).

PR #1114, opened 2022-08-30
juce::GenericInterpolator: Fixed wrong variable to assign to.

COx2:
I found a misstake of assigning to variable in juce::GenericInterpolator class.

Please consider correcting this with reference to PR.


reuk:
Thanks for reporting this issue. The problem seems to have been caused by copy-pasting code, so I've refactored a bit to reduce duplication:

https://github.com/juce-framework/JUCE/commit/d28815601a655407afe97100d68b4ab99c57600a

Issue #1118, opened 2022-09-05
[Bug]:  When there is only one item in a treeview bigger than the viewport and you scroll down it disappears


flyingrub:
### Detailed steps on how to reproduce the bug

Create a Treeview with only one item that is bigger than the viewport.
Scroll down.
The item disappear.
TreeView::ContentComponent::getAllVisibleItems does not return the correct results.

### What is the expected behaviour?

You can see the item scrolled down.

### Operating systems

Windows

### What versions of the operating systems?

Windows 10.

### Architectures

64-bit

### Testing on the `develop` branch

I have not tested against the `develop` branch

### Code of Conduct

- [X] I agree to follow the Code of Conduct

flyingrub:
```
        auto* item = [&]
        {
            auto* item = owner.rootItemVisible ? owner.rootItem
                                            : owner.rootItem->subItems.getFirst();

            auto* i = item;
            while (i != nullptr && i->y < visibleTop)
                i = getNextVisibleItem (i, true);

            return i == nullptr ? item : i;
        }();
```

reuk:
Thanks for reporting, that's fixed here: https://github.com/juce-framework/JUCE/commit/db90a31813f50b2d8bb65f12e80e95176f55805b

PR #1119, opened 2022-09-06
Fix scrolling when treeview only have one item bigger than the viewport

flyingrub:
Fixes https://github.com/juce-framework/JUCE/issues/1118


reuk:
Fixed in https://github.com/juce-framework/JUCE/commit/db90a31813f50b2d8bb65f12e80e95176f55805b

PR #1129, opened 2022-10-06
Fix GDI resource leak in WindowClassHolder

ValentynPavliuchenkoA:
Fix GDI resource leaks when using WindowClassHolder class: all icons returned by the ExtractAssociatedIcon() WinAPI function need to be destroyed by DestroyIcon().


reuk:
Fixed here: https://github.com/juce-framework/JUCE/commit/59a47a463873ad2a3515506ae983e430e2b6aba7

PR #1153, opened 2022-11-19
 Added ability to pass a Normalized Range to AudioParameterInt

DaniKog:
AudioParameterInt was not able to get a normalized range only min and max Added ability to pass it in


DaniKog:
I noticed I used this extensively in my local JUCE repro 
As always was making unique int ranges with a specific skew

Issue #1159, opened 2022-12-06
[Bug]: Raspberry Pi broken Graphics since 7.0.3 (tested on 64-bit)


benkuper:
### Detailed steps on how to reproduce the bug

Install fresh raspbian 64-bit on raspberry 4
clone JUCE 
make
launch Projucer

The window is buggy

Additional notes :

I’ve been compiling on CI for raspberry many times without problems, and I just discovered that one of my apps is having troubles rendering the main window on raspberry.
Strangely, I have another app that works just fine.
I decided to compile it directly from the Raspberry, starting with the Projucer, and I discovered that the Projucer itself is having the exact same problem

![image](https://user-images.githubusercontent.com/1291595/205895819-a801df0c-209a-4a70-af0c-2c77b500a125.png)

I’ll try to summarize my findings, you can try the bleeding edge version to see the problem

OS = Raspbian Bullseye 64-bit
JUEC : Fork synchronized to latest commit of develop branch (slightly modified but no big UI changes)
App1 = Chataigne ( [Chataigne | Chataigne 1](https://benjamin.kuperberg.fr/chataigne) )
App2 = LGML ( [LGML | Le Grand Mechant Loop](https://benjamin.kuperberg.fr/lgml) )

Normal GUI Mode (not using OpenGL Renderer) :
App1 = Works
App2 = Does not work (screenshot above)
Projucer = Does not work (screenshot above)

OpenGL renderer mode :
App1 = Works
App2 = Works but not really : the app shows quite alright but the menubar is showing the same bug
Projucer = Not tested with GL renderer

App1 & App2 are both using the same sub-framework for generating the UI ( OrganicUI : [GitHub - benkuper/juce_organicui: A massive JUCE module that adds multiple UI and control features 1](https://github.com/benkuper/juce_organicui) ) which allows to switch between renderers.

It’s quite reassuring that Projucer is having the same problem but then I don’t know why App1 is actually working…

If you have any idea about that I’m all ears !
Thank you

### What is the expected behaviour?

The GUI is showing normally

### Operating systems

Linux, Other

### What versions of the operating systems?

Raspbian OS 64-bit
JUCE 7.0.3 (latest develop branch)

### Architectures

ARM, 64-bit

### Testing on the `develop` branch

The bug is present on the `develop` branch

### Code of Conduct

- [X] I agree to follow the Code of Conduct

benkuper:
Small update on that : 
[...]

PR #1164, opened 2022-12-09
Fix: Linux GUIs (including Projucer) unresponsive

snystedt:
**Note that this issue also affects the pre-built Projucer from the webpage, making the GUI unresponsive on Linux.**

Just checked and the same solution has already been proposed in https://github.com/juce-framework/JUCE/issues/1159.

Well anyway, here's a PR with the same changes, and one extra which fixes an issue where the GUI won't start updating before an event is sent to the XWindow.

> This change fixes an issue where the GUI will not update at all if the refresh rate can't be determined. Since https://github.com/juce-framework/JUCE/commit/1da9ccd36c69d24af96d4fe65f9018026b8a80d6 only added getting the refresh rate from xrandr, any GUI application compiled with JUCE_USE_XRANDR=0 will not work. This change sets the refresh interval to 100 Hz (the same rate as before the vblank change) if the display object is missing the refresh rate.
> 
> Also added a call to updateVBlankTimer to avoid having to interact with the GUI in order to start the repaint timer.

### DIFF
--- a/modules/juce_gui_basics/native/juce_linux_Windowing.cpp
+++ b/modules/juce_gui_basics/native/juce_linux_Windowing.cpp
@@ -51,6 +51,7 @@ public:
             ++numAlwaysOnTopPeers;
 
         repainter = std::make_unique<LinuxRepaintManager> (*this);
+        updateVBlankTimer();
 
         windowH = instance->createWindow (parentToAddTo, this);
         parentWindow = parentToAddTo;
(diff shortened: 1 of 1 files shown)


reuk:
Thanks, this is fixed here: https://github.com/juce-framework/JUCE/commit/676e10fdd9e9641a258963f6b62003911005e656

PR #1168, opened 2022-12-18
Check pointer after casting in juce_ios_Audio.cpp dispatchAudioUnitPropertyChange

jamierpond:
I got a crash report from this function at some point last week. There might be something else going on, but I'm wondering if this is a sensible check anyway?


tpoole:
If `data` is nullptr then something else has very badly wrong and this check won't save you. It's clearest to indicate this by omitting the check.

PR #1170, opened 2022-12-21
Fix warnings in plugins in Xcode 14.2 when JUCE_VST3_CAN_REPLACE_VST2 is enabled

yairchu:
Relevant forum thread: https://forum.juce.com/t/plugins-no-longer-build-cleanly-with-xcode-14-1/53603


szarvas:
Thank you, the fix has been added in b997e72b62bb93fe8dd7801b00d43ee5152c7d01

PR #1177, opened 2022-12-24
Multichannel support for AudioFormatReader::searchForLevel

acendan:
`AudioFormatReader::searchForLevel` currently just checks the first two channels in a given file, regardless of channel count. The screenshots below show a simple command line utility I wrote, which checks for the first and last non-zero samples in an audio file, regardless of channel count. The bottom two files in the table include results from analyzing multi-channel files after making this change, with 8- and 10-channel files respectively. 

image
image


acendan:
For additional context, the current version would just stop at the second channel altogether and silently miss the non-zeros on any channels thereafter: 
image

acendan:
I've definitely filled out and signed the CLA... not sure why the check-CLA validation failed.

PR #1181, opened 2023-01-04
Fix projucer when building with clang under linux

dave-billin:
When the Projucer is built under Linux, the symbol "linux" must be undefined to avoid colliding with the TargetOS::OS::linux enumeration member.  Presently, this is only done when building with GCC.  This pull request modifies jucer_CommonHeaders.h to also under it when building with clang.

### DIFF
--- a/extras/Projucer/Source/Application/jucer_CommonHeaders.h
+++ b/extras/Projucer/Source/Application/jucer_CommonHeaders.h
@@ -25,11 +25,11 @@
 
 #pragma once
 
-
 //==============================================================================
-// The GCC extensions define linux somewhere in the headers, so undef it here...
-#if JUCE_GCC
- #undef linux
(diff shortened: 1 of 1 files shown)


reuk:
The build seems to work fine with clang on Fedora 37. I see that when building with GCC, the `linux` definition is provided by the compiler itself, but clang doesn't provide any such definition.

Perhaps you could try sticking a `#define linux 1` where we currently `#undef linux` and see if the compiler tells you the location of the previous definition. If it's in the standard library, it'd be helpful to know which OS + stdlib you're using. If it's a compiler builtin, it'd be helpful to know which compiler version you're using.

reuk:
Added here: https://github.com/juce-framework/JUCE/commit/c841b2ff2cd69e733d2a550f2cf5b69595857206

PR #1185, opened 2023-01-09
Fix Multicast on machines with multiple network interfaces

lukasklingebiel:
To receive Multicast packets on available network interfaces it's necessary to join the multicast group on all network interfaces individually. Currently it's not possible to control which network interface will receive multicast packets. On different macOS machines it can either be the WiFi or wired interface. This leads to issues on machines with multiple network interfaces.


medcelerate:
Can this be merged in, this would be very useful to my current app.

PR #1190, opened 2023-01-26
VST3: Fix bad value returned from setBusArrangements()

azeteg:
false == kResultTrue in Steinberg SDK, so this would cause a bad arrangement to be considered valid by host app.


reuk:
Thanks for the submission, that's added here: https://github.com/juce-framework/JUCE/commit/f7a28a6d624553f72d2c20811215f2937fc075e1

PR #1195, opened 2023-02-07
iOS change method of bypassing voice processing

snystedt:
The implementation of `AudioIODevice::setAudioPreprocessingEnabled` in `juce_ios_Audio.cpp` uses the `AVAudioSession::setMode` method to set the `AVAudioSessionMode` of the session to `AVAudioSessionModeMeasurement`. However, this also removes the processing of output signals, causing a drastic reduction in output playback level (see https://developer.apple.com/documentation/avfaudio/avaudiosessionmodemeasurement?language=objc). While this might be desirable behaviour for specific use cases, the `setAudioPreprocessingEnabled` description in `juce_AudioIODevice.h` states that

```
/** On devices which support it, this allows automatic gain control or other
    mic processing to be disabled.
    If the device doesn't support this operation, it'll return false.
*/
```

suggesting the intention is to only bypass audio **input** preprocessing.

My proposal is to instead use the `AudioUnitProperty` `kAUVoiceIOProperty_BypassVoiceProcessing` which only disables the input processing, not the output processing. (https://developer.apple.com/documentation/audiotoolbox/1534007-voice-processing_i_o_audio_unit_proper/kauvoiceioproperty_bypassvoiceprocessing?language=objc)

If this is not desirable, then I would recommend that another method be added, `AudioIODevice::setInputProcessingEnabled`, which would do the same thing suggested above.

This is crucial to record singing/vocals on iOS, since the Automatic Gain Control in the phones introduce a ton of noise when not actively singing/talking, as in when you're waiting for the place you want to come in with your singing.

### DIFF
--- a/modules/juce_audio_devices/native/juce_ios_Audio.cpp
+++ b/modules/juce_audio_devices/native/juce_ios_Audio.cpp
@@ -551,15 +551,10 @@ struct iOSAudioIODevice::Pimpl      : public AsyncUpdater
 
     bool setAudioPreprocessingEnabled (bool enable)
     {
-        auto session = [AVAudioSession sharedInstance];
+        const UInt32 one = 1;
+        AudioUnitSetProperty (audioUnit, kAUVoiceIOProperty_BypassVoiceProcessing, kAudioUnitScope_Input, 1, &one, sizeof (one));
 
-        NSString* mode = (enable ? AVAudioSessionModeDefault
-                                 : AVAudioSessionModeMeasurement);
-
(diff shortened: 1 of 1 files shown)


snystedt:
Alright, so this was kind of a fast PR I made when discovering the issues with `AVAudioSessionModeMeasurement`. I'll come back to this and clean up the code. However it would be good to know if this is something that you'd consider merging in near-time. Currently I have to hack a modified `juce_ios_audio.cpp` into our build process and would obviously prefer to stop doing so asap. :)

PR #1202, opened 2023-02-20
fix: Use atomic in opengl to avoid race

el-type:
I'm not sure if my company has signed the contribution document, but I would assume so.
I don't think my user name or email was a part of it though.


reuk:
I believe this issue should be addressed by 42021e3e9a237f3ac19242560605496a69e90324. Please let us know if you're still seeing data races on develop.

PR #1224, opened 2023-04-27
Allow noise gate to work as an expander

maxwalley:
The noise gate documentation says that if the ratio is low enough it can be used as an expander, however, there is an assertion forcing the ratio to be above or equal to 1.0. This stops the gate being usable as an expander. I have simply removed this assertion and edited the documentation.


modules/juce_dsp/widgets/juce_NoiseGate.cpp
```diff
@@ -49,8 +49,6 @@ void NoiseGate<SampleType>::setThreshold (SampleType newValue)
 template <typename SampleType>
 void NoiseGate<SampleType>::setRatio (SampleType newRatio)
 {
-    jassert (newRatio >= static_cast<SampleType> (1.0));
```
jamierpond, line 52: Possibly worthwhile checking for negative ratios?

jamierpond (approved):

PR #1227, opened 2023-05-04
Add support for EmbeddedFrameworks in custom juce modules 

benkuper:
This adds a property line for adding embedded frameworks for xcode exporters from a custom juce module header file.
The JUCE Module Format documentation is also updated inside this PR


jameshball:
Just wanted to +1 this PR as this would be super useful for juce_sharedtexture

PR #1239, opened 2023-06-15
Add support for the musl libc.

vimproved:
JUCE currently uses multiple glibc-specific features when JUCE_LINUX is defined. Instead, this pull request adds two new macros in juce_TargetPlatform.h: JUCE_GLIBC and JUCE_MUSL. JUCE_GLIBC is defined if the system is using glibc (detected through the __GLIBC__ macro which should be defined on all glibc systems), and JUCE_MUSL is defined otherwise. This pull request also conditions the usage of execinfo.h (only available on glibc), the usage of the stat64 interface (being removed in the next release of musl, currently only available when _LARGEFILE64_SOURCE is defined), and the usage of _NL_ADDRESS_LANG_AB and _NL_ADDRESS_COUNTRY_AB2 (locales are not supported by musl) behind JUCE_GLIBC instead of JUCE_LINUX.

### DIFF
--- a/modules/juce_core/juce_core.cpp
+++ b/modules/juce_core/juce_core.cpp
@@ -102,7 +102,7 @@
  #include <net/if.h>
  #include <sys/ioctl.h>
 
- #if ! (JUCE_ANDROID || JUCE_WASM)
+ #if ! (JUCE_ANDROID || JUCE_WASM || JUCE_MUSL)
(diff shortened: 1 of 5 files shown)


vimproved:
not sure if JUCE even checks pull requests, but I'll revive this for the sake of packagers and people using their own forks. will rebase at some point

PR #1246, opened 2023-07-18
Add quotes to Projucer generated Makefiles

FigBug:
If plugins have spaces in their names they are currently failing to build.


reuk:
Should be resolved by https://github.com/juce-framework/JUCE/commit/3ea663e91ee6d3b1aec456515ee17f856ecce842

PR #1249, opened 2023-07-28
[Bugfix] Add missing early return in AudioSubsectionReader::readSamples()

danra:
Looks like a very old regression since 27895cb6bd49dbe493f136801e9c43ee8c5f366f

In case `clearSamplesBeyondAvailableLength()` returns a negative number of samples, we need to avoid calling `readSamples()` on the source reader. One reason is that it's possible the source reader `readSamples()` will also call `clearSamplesBeyondAvailableLength()`, which can end up calling `zeromem()` with that negative number of samples, causing a crash. (that would happen in case `samplesAvailable`'s value would be an even lower negative number, causing the condition `(samplesAvailable < numSamples)` to evaluate to true).


danra:
Ping. 
This fixes an actual crash scenario, and would be helpful to merge in for all JUCE users.

Anthony-Nicholls:
This is fixed in https://github.com/juce-framework/JUCE/commit/de78cd0c065008b386bf25208f7db052e0d2ce33. 

Thanks for reporting.

PR #1279, opened 2023-09-26
AudioProcessor // Stop crashing when debugging standalone in Xcode.

ShikiSuen:
This PR bypasses AU host identity validation when running in Xcode debug mode.


reuk:
This behaviour is intentional, to alert users that they need to update their code.

PR #1281, opened 2023-09-28
fix: abstract FIFO example docs

ncblair:
in the addToFifo method, the second copySomeData should start copying from someData + scope.blockSize1 instead of someData


tpoole:
Thank you. Added here: https://github.com/juce-framework/JUCE/commit/65f55e17acecf76a12022a704e1f16857128d593

PR #1296, opened 2023-10-30
Crash fix: Remove Concertina header mouse listener on destroy

emezeske:
This fixes a crash bug where the header component (provided that it was owned by the caller, i.e. shouldTakeOwnership=false) has a dangling mouse listener pointing to "this," which is a PanelHolder that has been destroyed. We need to remove this listener when either the PanelHolder is destroyed, or a new custom header has been set. Otherwise, if that custom header is used again, mousing over it will SIGSEGV when the mouse listener is invoked.

### DIFF
--- a/modules/juce_gui_basics/layout/juce_ConcertinaPanel.cpp
+++ b/modules/juce_gui_basics/layout/juce_ConcertinaPanel.cpp
@@ -223,6 +223,11 @@ public:
         addAndMakeVisible (comp);
     }
 
+    ~PanelHolder() {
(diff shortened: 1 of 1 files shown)


reuk:
Thanks for reporting. This should be fixed in these commits:

https://github.com/juce-framework/JUCE/commit/884bb756c873575cd651310e333cf5721c84d857
https://github.com/juce-framework/JUCE/commit/65b6c85594ecb98645de5048e7b496cc84c23659

emezeske:
Woohoo, thanks, @reuk!

Issue #1315, opened 2023-12-12
[Bug]: Modals cancelled when app is minimised


quickiwiki:
### Detailed steps on how to reproduce the bug

I have seen an odd behaviour in the model component manager (ModalComponentManager::ModalItem) that doesn’t seem to be right and causes the modals to be cancelled when the JUCE application is minimised. At its core, componentVisibilityChanged() is called when the application is minimised, and it looks like this:

void componentVisibilityChanged() override
{
    if (! component->isShowing())
        cancel();
}

The code above would cancel any modal. 

### What is the expected behaviour?

Modal windows should not be cancelled on minimise.

Such behaviour can be fixed with an easy change, which might resemble the code below.

void componentVisibilityChanged() override
{
    const auto peer = component->getPeer();
    if (! component->isShowing() && !(peer && peer->isMinimised()))
        cancel();
}

### Operating systems

macOS

### What versions of the operating systems?

macOS Monterey 12.6.7

### Architectures

ARM

### Testing on the `develop` branch

The bug is present on the `develop` branch

### Code of Conduct

- [X] I agree to follow the Code of Conduct

szarvas:
Thank you for reporting. While the proposed change is a fix, we opted to go for a different change. The general thinking behind the cancel() implementation, is that the user interface shouldn't become unresponsive due to a modal component in a way, that is not easily visible/discoverable by the user.

The Windows implementation sidesteps this issue, by ignoring close and minimise request during modal activity. A change has been released on `develop`, which mimicks the Windows behaviour.

https://github.com/juce-framework/JUCE/commit/fb14118771eeb34f51146e23def0a1f94a49fbad

PR #1316, opened 2023-12-12
Fix Modals cancelled when app is minimised

quickiwiki:
This PR resolves issue https://github.com/juce-framework/JUCE/issues/1315

### DIFF
--- a/modules/juce_gui_basics/components/juce_ModalComponentManager.cpp
+++ b/modules/juce_gui_basics/components/juce_ModalComponentManager.cpp
@@ -52,7 +52,8 @@ struct ModalComponentManager::ModalItem final : public ComponentMovementWatcher
 
     void componentVisibilityChanged() override
     {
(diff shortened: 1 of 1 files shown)


stephenk-az:
This was the same fix I suggested here (essentially) and attila provided a good reason why this would NOT be the best fix:

https://forum.juce.com/t/bug-minimize-to-dock-on-macos-causes-modal-component-to-lose-modality/57974/5

"The suggested workaround could break existing use-cases. In an application with multiple windows, minimising the one with the modal component would leave the visible window in a non-interactive state, and it would be difficult to discover why this is the case."

szarvas:
A change has been released on `develop`

https://github.com/juce-framework/JUCE/commit/fb14118771eeb34f51146e23def0a1f94a49fbad

The minimise and close buttons are now disabled when a component is modal, to avoid this problem the same way Windows does.

PR #1332, opened 2024-01-23
Allow daisy-chaining when adding children to ValueTrees

ImJimmi:
Currently, it's possible to daisy-chain property setters on a value tree:

```cpp
juce::ValueTree {"Tree"}
    .setProperty("foo", 10, nullptr)
    .setProperty("bar", 20, nullptr);
```

This PR adds this behaviour to the child-adders as well, allowing a whole tree to be constructed this way:

```cpp
juce::ValueTree {"Root"}
    .setProperty("foo", 10, nullptr)
    .setProperty("bar", 20, nullptr)
    .appendChild (juce::ValueTree {"Branch"}
                      .setProperty ("x", y, nullptr)
                      .appendChild (juce::ValueTree {"Leaf"}, nullptr),
                  nullptr)
    .addChild (juce::ValueTree {"Branch"}, 0, nullptr);
```

This is useful when loading value-trees from places external to the current call site (.e.g loading state from XML) and then adding on any additional children (e.g. any state that shouldn't be persisted to a file).

It also allows for better const-ness as the tree doesn't need to be changed after initialisation:

```cpp
juce::ValueTree nonConst {"Tree"};
nonConst.appendChild (juce::ValueTree {"Data"}, nullptr);

const auto isConst = juce::ValueTree {"Tree"}
                         .appendChild (juce::ValueTree {"Data"}, nullptr);
```

### DIFF
--- a/modules/juce_data_structures/values/juce_ValueTree.cpp
+++ b/modules/juce_data_structures/values/juce_ValueTree.cpp
@@ -244,7 +244,7 @@ public:
         return children.indexOf (child.object);
     }
 
-    void addChild (SharedObject* child, int index, UndoManager* undoManager)
+    ValueTree& addChild (SharedObject* child, int index, UndoManager* undoManager)
     {
         if (child != nullptr && child->parent != this)
         {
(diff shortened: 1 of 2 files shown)


ImJimmi:
This would go nicely with https://github.com/juce-framework/JUCE/pull/1333 to make constructing value trees even cleaner:

```cpp
juce::ValueTree {"Root"}
    .setProperty("foo", 10)
    .setProperty("bar", 20)
    .appendChild (juce::ValueTree {"Branch"}
                      .setProperty ("x", y)
                      .appendChild (juce::ValueTree {"Leaf"}))
    .addChild (juce::ValueTree {"Branch"}, 0);
```

PR #1340, opened 2024-01-28
Documentation: Remove duplicate apt dependency

paulgration:
- Removes duplicate apt dependency (`libxcursor-dev`)
- Splits apt dependencies onto new lines to make it easier to spot duplicates


tpoole:
Thank you for reporting.

Added here: https://github.com/juce-framework/JUCE/commit/51edcaf842e0ae281de78ce33dd998846897a710

We have kept multiple packages on single lines to group related dependencies.

PR #1352, opened 2024-02-15
Changed how HTTP headers are ended on Linux

maxwalley:
This is fixing the error reported at https://forum.juce.com/t/url-header-suffix-on-linux/59985/2

Essentially the HTTP header section should end with "/r/n/r/n". The previous code was not taking into account that there might already be some of the suffix present before applying it.

### DIFF
--- a/modules/juce_core/native/juce_Network_linux.cpp
+++ b/modules/juce_core/native/juce_Network_linux.cpp
@@ -525,7 +525,14 @@ private:
         if (userHeaders.isNotEmpty())
             header << "\r\n" << userHeaders;
 
-        header << "\r\n\r\n";
+        if(header.toString().endsWith("\r\n"))
+        {
+            header << "\r\n";
+        }
+        else
+        {
+            header << "\r\n\r\n";
+        }
 
         if (hasPostData)
             header << postData;


modules/juce_core/native/juce_Network_linux.cpp
```diff
@@ -525,7 +525,14 @@ class WebInputStream::Pimpl
         if (userHeaders.isNotEmpty())
             header << "\r\n" << userHeaders;
 
-        header << "\r\n\r\n";
+        if(header.toString().endsWith("\r\n"))
```
aminya, line 528: I am sure there's a better to detect `\r\n` than creating a string copy here.

maxwalley: I have added another commit which uses a reference to a variable.

aminya: What I meant was the `toString` call, which I think creates a string copy. The actual string literals were fine. 
It should not matter much as headers should be small.

maxwalley: The header variable is a juce::MemoryOutputStream. I do not know how to check the end of that for the string I am looking for but I would argue that my method of fixing it is better than having the bug present. This breaks our licensing in our application and we have to used a forked version instead.

sevoster:
so what's the progress on this PR? We face the same issue on Linux that JUCE sends request with extra /r/n and it breaks parsing of request on our server

maxwalley:
We currently have had to move to a forked version of JUCE found here: https://github.com/salsa-sound/JUCE-6

This request has now been open for over a year with no resolution.

PR #1360, opened 2024-02-29
Add Energy to StatisticsAccumulator

cjappl:
Add the energy (sum of squares) to the output of the StatisticsAccumulator.


cjappl:
A couple side notes:
1. I noticed there don't seem to be unit tests for this class, or I just missed them. Could you point me to them if they exist and I'll add a few tests for this?
2. If this is acceptable, I could add other statistics that seem to be missing (power, and RMS being the major ones). Any interest in me following up with that?

PR #1370, opened 2024-03-29
Fix crash after MessageManager deleted/recreated

emezeske:
If the MessageManager is deleted (via e.g. deleteInstance()) and then re-created in the same process, the next time that a top-level window is created, the process will crash because the JuceIVirtualDesktopManager COM object pointer is cached in a static variable, and the MessageManager deconstructor calls OleUninitialize() which shuts down COM, thereby invalidating the static cached pointer.

There's no reason to cache this anyway, as CoCreateInstance() returns a singleton, and this is not performance-sensitive code in a tight loop or something.

I found this bug because in my unit testing, I create and destroy the message manager for each test to ensure that the tests are hermetic. With this fix in place it works perfectly.

Thank you for submitting a pull request.

### DIFF
--- a/modules/juce_gui_basics/native/juce_Windowing_windows.cpp
+++ b/modules/juce_gui_basics/native/juce_Windowing_windows.cpp
@@ -5336,19 +5336,14 @@ bool juce::detail::WindowingHelpers::isWindowOnCurrentVirtualDesktop (void* x)
     if (x == nullptr)
         return false;
 
-    static auto* desktopManager = []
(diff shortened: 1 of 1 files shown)


reuk:
Thanks for reporting, we've made a similar change here: https://github.com/juce-framework/JUCE/commit/c0d0c1f9c6ed3a37d213c2f651c8b97a506cffc5

emezeske:
Thanks!

PR #1385, opened 2024-05-30
ValueTree::SharedObject::callListeners optimization

DamRsn:
This PR introduces an optimization to the `ValueTree::SharedObject::callListeners` method, improving its performance by avoiding a `valueTreesWithListeners.contains(v)` call when `valueTreesWithListeners` is not modified by one of the listeners. Instead, `valueTreesWithListeners[i] == v` is checked before. It gives a significant speedup when `numListeners` is high.

Note that if `i` is out of bounds, a `nullptr` is returned by `valueTreesWithListeners[i]` and it will compare as not equal to `v`.

This change improved significantly the performance of preset loading at Minimal Audio (internally using `juce::ValueTree::copyPropertiesAndChildrenFrom`), so we share it back. See this [PR description in our JUCE fork](https://github.com/Minimal-Audio/JUCE/pull/7) for some indicative benchmarks.

### DIFF
--- a/modules/juce_data_structures/values/juce_ValueTree.cpp
+++ b/modules/juce_data_structures/values/juce_ValueTree.cpp
@@ -81,7 +81,7 @@ public:
             {
                 auto* v = listenersCopy.getUnchecked (i);
 
-                if (i == 0 || valueTreesWithListeners.contains (v))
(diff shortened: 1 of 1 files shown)


Anthony-Nicholls:
@DamRsn thanks for sharing, I've pushed a change here https://github.com/juce-framework/JUCE/commit/315167a91a42a520d0a6d3e9fe63f913583329ef

PR #1396, opened 2024-06-22
Add Support for OSC Types int64, double, Nil, Impulse, T and F

benkuper:
This PR adds support for both OSC receiver and sender for the types int64 (h), double (d), Nil (N), Impulse (I), True (T) and False (F).

Many software use those types (especially T and F), and without proper support for it, it is impossible to receive messages from those, rendering the module quite useless.

UnitTest have been updated to include those types.

_Note : I already had made a PR years ago for this, that is has been ignored I believe. With Juce 8 releasing, I'm taking again my chance to resync my repo as much as possible with the current one, making neat PR for everything that I changed in my fork and that I believe will benefit to many other developers.
Please, consider at least checking it and replying, thank you and have a nice day_

### DIFF
--- a/modules/juce_osc/osc/juce_OSCArgument.cpp
+++ b/modules/juce_osc/osc/juce_OSCArgument.cpp
@@ -36,10 +36,15 @@ namespace juce
 {
 
 OSCArgument::OSCArgument (int32 v)              : type (OSCTypes::int32),   intValue (v) {}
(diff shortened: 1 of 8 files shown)


benkuper:
I hate insisting, but please, pleaaase, this (and the other PR I made) or really straightforward and would just bring the OSC support up to an actual functional state, otherwise it's just an idea of something that would be interesting..

Issue #1377, opened 2024-04-11
[Bug]: Switch from webkit2gtk 4.0 to 4.1


jbicha:
### Detailed steps on how to reproduce the bug

webkitgtk introduced the 4.1 API more than a year ago. It is the same as the 4.0 API but uses libsoup3 instead of libsoup2.4. Since JUCE doesn't use libsoup directly, it is an easy swap to switch to the 4.1 API. Every stable Linux distribution with a supported version of webkitgtk already provides the 4.1 packages.

More importantly, Linux distributions have begun removing the 4.0 packages. This is true for Debian Unstable, Ubuntu 24.04 LTS, and probably Fedora 41 (to be released later this year).

### What is the expected behaviour?

A command like this can do the switch:

```
find . -type f -print0 | xargs -0 sed -i '' -e 's/webkit2gtk-4.0/webkit2gtk-4.1/g'
```

### Testing on the `develop` branch

The bug is present on the `develop` branch

### Code of Conduct

- [X] I agree to follow the Code of Conduct

KottV:
It become more reasonable with this commit https://github.com/juce-framework/JUCE/commit/2516ad808efad4ce27ce6070dfe0c4009742356f
As I can see we need to set `webkit2gtk-4.1` as default in https://github.com/juce-framework/JUCE/blob/61a03097ec9e01693c87ac71935e97b9714cff1a/extras/Build/CMake/JUCEUtils.cmake#L98 and https://github.com/juce-framework/JUCE/blob/61a03097ec9e01693c87ac71935e97b9714cff1a/extras/Projucer/Source/ProjectSaving/jucer_ProjectExporter.cpp#L614

jbicha:
Ok, I'll close the bug now.

KottV:
> Ok, I'll close the bug now.

I meant that it's still an issue, and with the mentioned commit  things are inconsistent, JUCE partially wants 4.1 and partially 4.0.

szarvas:
We have a fix for this issue shortly coming up.

PR #1404, opened 2024-06-25
Dependency Update: webkit2gtk4.0 -> webkit2gtk4.1

tdug:
Addresses https://github.com/juce-framework/JUCE/issues/1377

I'm not knowledgeable enough with C++/CMake for it to be obvious to me how to allow either version of this dependency. Would love some feedback if this is possible so I don't break any builds that cannot install webkit2gtk-4.1 for some reason.

This was a simple find&replace. I'm not sure if there's anything else to update here.


docs/Linux Dependencies.md
davidhealey, line 1: According to this very doc v4.1 is already the standard as it says on older systems you can use 4.0.

modules/juce_gui_extra/native/juce_WebBrowserComponent_linux.cpp
davidhealey, line 1: I think you need to revisit this PR. You can see in the code below that it is already pulling v4.1 with v4.0 as a fallback, but your change eliminates the fallback and doesn't seem to have any benefit.

tdug: Oh you're right. I just did a naive find and replace here. I'll update this on my next revision.

davidhealey: I've actually been digging into this myself in a JUCE 6 based project today. I found I also needed to swap some code out in the jucer_ProjectExporter.cpp file, not sure if it's relevant in the latest JUCE but I'll leave it here in case you need it.

```
    if (isWebBrowserComponentEnabled (project) && type == PackageDependencyType::compile)
    {
		if (library.open ("libwebkit2gtk-4.1.so"))
			packages.add ("webkit2gtk-4.1");
        else
			packages.add ("webkit2gtk-4.0");
			
        packages.add ("gtk+-x11-3.0");
    }
   ```

tdug:
@davidhealey - I appreciate your feedback on this PR. I think this feature is included in a more pragmatic way in the latest `develop` branch. I'm closing this PR because of this.

PR #1405, opened 2024-06-25
GitHub workflow improvements

tdug:


.github/workflows/check-cla.yml
```diff
@@ -1,32 +1,24 @@
-name: check-CLA
-on: [pull_request_target]
+name: Check CLA
+
+on:
+  pull_request:
```
tdug, line 4: This could be changed in a future PR to support `pull_request_target`.

.github/workflows/check-cla.yml
```diff
@@ -1,32 +1,24 @@
...
+
 jobs:
   check-cla:
     runs-on: ubuntu-latest
+    continue-on-error: ${{ github.repository_owner != 'juce' }}
```
tdug, line 9: This allows this step to be skipped in forks.

PR #1436, opened 2024-10-04
[ENG-2008] JUCE 8 Update

ancientjpeg:
## Notes

- Seems like our optimization from #7 was incorporated into the JUCE lib.


ancientjpeg:
Apologies, this was meant to be opened on our fork.

PR #1439, opened 2024-10-08
reFX:   Projucer:   Launch correct VS version. Don't launch again if already running.

FigBug:
Fixes the 2 most annoying ProJucer bugs on Windows

1. If you click "Save and Open in IDE" often the wrong version of Visual Studio will launch
2.  If you click "Save and Open in IDE" and Visual Studio is already running, it will open a second copy instead of just letting Visual Studio reload the existing project.

This change brings the behaviour of the Windows ProJucer in line with the macOS version.

### DIFF
--- a/extras/Projucer/Source/ProjectSaving/jucer_ProjectExport_MSVC.h
+++ b/extras/Projucer/Source/ProjectSaving/jucer_ProjectExport_MSVC.h
@@ -1590,6 +1590,13 @@ public:
     bool launchProject() override
     {
        #if JUCE_WINDOWS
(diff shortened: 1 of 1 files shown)


reuk:
Thanks for the suggestion.

The change to avoid opening duplicate instances of VS was added in 04a04cffcdc930b84a068771647a51527ee216d7.

To open the correct version of VS, right-click a solution file and select "Properties". Under "Open with" select "Change", and select the Microsoft Visual Studio Version Selector. Now, sln files will always open with the appropriate version of Visual Studio.

Issue #1445, opened 2024-10-11
[Bug]: [Juce8 / QuickJS] Random but inevitable script stopping with error "interrupted"


benkuper:
### Detailed steps on how to reproduce the bug

I have this very simple script, the function "update" is called in a dedicated thread at 100fps.

```
function update(deltaTime) {
	for (var i = 0; i < 100; i++) {

	}
}
```

Eventually, this code will stop, the returned error code being "interrupted".

This is the stack trace when breaking at the throwing of this interruption :

```
>	Chataigne.exe!`anonymous namespace'::choc::javascript::quickjs::__js_poll_interrupts(`anonymous-namespace'::choc::javascript::quickjs::JSContext * ctx) Line 16707	C++
 	Chataigne.exe!`anonymous namespace'::choc::javascript::quickjs::js_poll_interrupts(`anonymous-namespace'::choc::javascript::quickjs::JSContext * ctx) Line 16718	C++
 	Chataigne.exe!`anonymous namespace'::choc::javascript::quickjs::JS_CallInternal(`anonymous-namespace'::choc::javascript::quickjs::JSContext * caller_ctx, `anonymous-namespace'::choc::javascript::quickjs::JSValue func_obj, `anonymous-namespace'::choc::javascript::quickjs::JSValue this_obj, `anonymous-namespace'::choc::javascript::quickjs::JSValue new_target, int argc, `anonymous-namespace'::choc::javascript::quickjs::JSValue * argv, int flags) Line 27286	C++
 	Chataigne.exe!`anonymous namespace'::choc::javascript::quickjs::JS_CallFree(`anonymous-namespace'::choc::javascript::quickjs::JSContext * ctx, `anonymous-namespace'::choc::javascript::quickjs::JSValue func_obj, `anonymous-namespace'::choc::javascript::quickjs::JSValue this_obj, int argc, `anonymous-namespace'::choc::javascript::quickjs::JSValue * argv) Line 28655	C++
 	Chataigne.exe!`anonymous namespace'::choc::javascript::quickjs::JS_Invoke(`anonymous-namespace'::choc::javascript::quickjs::JSContext * ctx, `anonymous-namespace'::choc::javascript::quickjs::JSValue this_val, unsigned int atom, int argc, `anonymous-namespace'::choc::javascript::quickjs::JSValue * argv) Line 28809	C++
[...]

PR #1448, opened 2024-10-17
Fix javascript bad interrupt

benkuper:
This proposes a fix for the issue #1445 
 
The execute() function sets an interrupt in the engine that is then checked whenever a function is called using callFunction, but the interval in the interrupt check function is calculated depending on the time at execute(), not the time at callFunction(), resulting in an inevitable interrupt if we keep calling functions after having calling execute() once.

This fix store a "timeAtLastStart" variable that replaces the lambda-local "started" argument. This way, it is effectively updated everytime an evaluate / execute / callFunction is used, without having to redefine the engine interrupt handler everytime.

A more complete and possibly elegant implementation would be to keep track of which thread has called what and have different checks for each thread, so they can be treated independently. But this fix is way simpler and should be good for most cases.

### DIFF
--- a/modules/juce_core/javascript/juce_Javascript.cpp
+++ b/modules/juce_core/javascript/juce_Javascript.cpp
@@ -719,10 +719,10 @@ public:
         return var::undefined();
     }
 
-    Result execute (const String& code, RelativeTime maxExecTime)
+    Result execute (const String& code, RelativeTime maxExecutionTime)
     {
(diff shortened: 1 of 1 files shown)


reuk:
This issue was fixed here: https://github.com/juce-framework/JUCE/commit/3cca812d3804b3dbc729685d068428b25b384d3d

PR #1459, opened 2024-10-24
Fixed TextLayout::getStringBounds nullptr error for empty strings

FangCunWuChang:
Calling TextLayout::getStringBounds() on an empty string will result in a null pointer error.


tpoole:
Thank you.

Added here: https://github.com/juce-framework/JUCE/commit/a784d5776be6daef574cc5b6fe79b182c0e4dc42

PR #1461, opened 2024-10-24
Fixed classes symbol export

FangCunWuChang:
Some ARA Host classes do not add the JUCE_API macro, which makes it impossible to export these symbols when compiling JUCE into a dynamic link library, resulting in the generated dynamic link library being unable to be linked.
These are some symbols other than those fixed by the previous PR #1454 .


reuk:
Thanks, that's merged here: https://github.com/juce-framework/JUCE/commit/d9ed81b4f039dfd3bae2ff51fa4f949f935fe79b

PR #1468, opened 2024-11-07
Update MidiMessageSequence::updateMatchedPairs to deal with note-on messages with velocity 0

FangCunWuChang:
In the current JUCE, both the `juce::MidiMessage::isNoteOn` and `juce::MidiMessage::isNoteOff` methods allow users to decide how to regard notes with velocity 0. Here, I have also added this option to the `juce::MidiMessageSequence::updateMatchedPairs` method.

When `regardNoteOnEventsWithVel0AsNoteOff` is set to `true` (the default), it will maintain the original behavior of regarding note-on events with velocity 0 as note-off events.


FangCunWuChang:
Sorry, deleted remote branch by mistake.

PR #1476, opened 2024-11-19
Code analysis issues

Dave-Lowndes:
assert opcode before it's used.
Don't fail to delete memory even though it's unlikely to occur.
Handle situation when SHBrowseForFolder is cancelled (returns null).
Several places appear to be checking the wrong thing after calling SafeArrayCreateVector.


reuk:
Please can you provide some details about the code analysis you're running? Which tool are you using, and with what settings?

Dave-Lowndes:
@reuk Visual Studio code analysis & PVS-Studio - default settings. These are just a few that I've picked on while going over the Surge XT sources.

PR #1478, opened 2024-11-21
Fix UB in VST plugins due to function prototype mismatch

omartijn:
This fixes undefined behaviour being triggered by the fact that the function signatures used for functions in VST plugins do not match those defined in the VST SDK.

Now that that's fixed some function pointer casts could also be removed (since the function types now actually match), so I've done that as well.


reuk:
Resolved by the following patches:

https://github.com/juce-framework/JUCE/commit/6e910d801050e017a590a7ac2d0d2f3ea0bd187f
https://github.com/juce-framework/JUCE/commit/73cb3b88adbce84abd30ed2bef765278241d2ca1

PR #1516, opened 2025-02-27
Fix: Potential Vulnerability in Cloned Function

tabudz:
**Description**
This PR fixes a security vulnerability in crc32_big() that was cloned from zlib but did not receive the security patch. The original issue was reported and fixed under https://github.com/madler/zlib/commit/d1d577490c15a0c6862473d7576352a9f18ef811.
This PR applies the same patch to eliminate the vulnerability.

**References**
https://nvd.nist.gov/vuln/detail/CVE-2016-9843
https://github.com/madler/zlib/commit/d1d577490c15a0c6862473d7576352a9f18ef811


reuk:
Thanks, we've now updated to zlib 1.3.1, which includes this fix.

PR #1581, opened 2025-10-13
Fix Text Editor’s caret being set incorrectly for trailing whitespaces

aggniesz:
When trying to move Text Editor’s caret to the end of the last line of the text, it was being set one character to the left of the last whitespace character. 

getGlyphIndexToTheRightOf is now reporting an invalid index for trailing whitespaces (as there’s nothing to the right of it), and the last character’s index for newline characters.

Also fixes newline characters not being detected correctly due to removing all control characters during string sanitising in SimpleShapedText’s constructor.


reuk:
Thanks, I've pushed these changes with a few tweaks:

https://github.com/juce-framework/JUCE-dev/commit/3a0135ffb7f2bfe05a99f9b4daf89beba7e4d7eb

https://github.com/juce-framework/JUCE-dev/commit/e525e12061a29530acc11f6e4c6a25268f869483

PR #1597, opened 2025-11-24
Correct VST3_CATEGORIES defaults in CMake API docs

szkkng:
This change updates the CMake API documentation so that the default VST3_CATEGORIES mention both Instrument and Synth when IS_SYNTH is TRUE.

According to the _juce_set_fallback_properties function in extras/Build/CMake/JUCEUtils.cmake, when IS_SYNTH is TRUE the default VST3_CATEGORIES are Instrument and Synth:

```cmake
if(is_synth)
    _juce_set_property_if_not_set(${target} VST3_CATEGORIES Instrument Synth)
else()
    _juce_set_property_if_not_set(${target} VST3_CATEGORIES Fx)
endif()
```

However, the CMake API documentation currently mentions only Synth as the default in this case.
This change updates the docs so that the described defaults match the implementation.


reuk:
Merged in https://github.com/juce-framework/JUCE/commit/230340dbfc580628596243bb63b31aeb4d6b5a6d, thanks!

PR #1607, opened 2025-12-16
Fix: Allow nullptr callback in evaluateJavascript on Linux

kinoshita-lab:
The callback parameter of WebBrowserComponent::evaluateJavascript has a default value of nullptr. Updated the Linux implementation to handle this case properly by checking if the callback queue is empty before attempting to invoke the callback.

Both Windows and macOS implementations already allow nullptr for the callback parameter. This change makes Linux work the same way.


tpoole:
Thank you for reporting. We've added a fix here: https://github.com/juce-framework/JUCE/commit/f6a81fe5f29cc7e4d4a8de07baeffe1994996f98

PR #1617, opened 2026-01-25
Fix VST3_PATH and LV2_PATH not being respected

### LINKED ISSUE #1189: [Bug]: Honor LV2_PATH/VST3_PATH

Start a JUCE program, e.g. BespokeSynth: plugins listed in `VST3_PATH` are not taken into account.

The PATH to search for LV2/VST3 should honor the `LV2_PATH`/`VST3_PATH` environment path (colon separated list of folders to search for the VST plugins). This is already done for the `LADSPA` format [here](https://github.com/juce-framework/JUCE/blob/4e68af7fde8a0a64e808fb1a37999a690873c9de/modules/juce_audio_processors/format_types/juce_LADSPAPluginFormat.cpp#L715) and VST [here](https://github.com/juce-framework/JUCE/blob/4e68af7fde8a0a64e808fb1a37999a690873c9de/modules/juce_audio_processors/format_types/juce_VSTPluginFormat.cpp#L3645) so it should not be to hard to modify (actually any reasons why on MacOS you don't read this variable as well?):
- [this code](https://github.com/juce-framework/JUCE/blob/4e68af7fde8a0a64e808fb1a37999a690873c9de/modules/juce_audio_processors/format_types/juce_LV2PluginFormat.cpp#L5267-L5289) for LV2 plugins (to read `LV2_PATH`)
-  [this code](https://github.com/juce-framework/JUCE/blob/4e68af7fde8a0a64e808fb1a37999a690873c9de/modules/juce_audio_processors/format_types/juce_VST3PluginFormat.cpp#L3941) for the `VST3_PATH`.

### Motivation

Plugins can be located in various places on the system. While LV2 [proposes](https://github.com/lv2/lv2) `LV2_PATH` to search for the paths, [the spec](https://steinbergmedia.github.io/vst3_dev_portal/pages/Technical+Documentation/Locations+Format/Plugin+Locations.html) for VST3 only mentions fixed paths: unfortunately, sometimes these paths can be impossible to use, either because the OS does not provide such path for security or purity reasons (e.g. Flatpack or NixOs) or because the programs are run without installing them (e.g. nix provides `nix run` that can run a program without installing it, and I know some people trying to build around it a whole linux music portable installation this way, like https://codeberg.org/PowerUser/universal-studio that actually bundles BespokeSynth… except that it can't detect the plugins for the reasons explained here).

Multiple people reported this as an issue [here](https://github.com/steinbergmedia/vst3sdk/issues/67), and some of them started to use `VST3_PATH` including [Ardour and Carla](https://github.com/flathub/org.freedesktop.LinuxAudio.BaseExtension/issues/5#issuecomment-820822234). It would be great to see BespokeSynth here as well! 

Related: https://github.com/BespokeSynth/BespokeSynth/issues/1063

### Operating systems

macOS, Linux

All of them (tested on NixOs, but the code is explicit)

### Architectures

x86_64, ARM, Other, 64-bit, 32-bit

### Stacktrace

_No response_

VST3, LV2

All (tested in Bespokesynth)

The bug is present on the `develop` branch

### Code of Conduct

hcsch:
Use the content of the environment variables as the default search path,
as is already the case for VST and LADSPA.

Re-order paths for LV2 and VST3 so that more local / more user-specific
paths come earlier in the search paths than more global / more
system-wide paths.
For LV2 this matches the recommendations on
https://lv2plug.in/pages/filesystem-hierarchy-standard.html#:~:text=Bundle%20Discovery

Existing search path order seems to be inconsistent across plugin
formats and OSes.

Fixes https://github.com/juce-framework/JUCE/issues/1189

I've tested this to work (with JUCE as a dependency of BespokeSynth)

### DIFF
--- a/modules/juce_audio_processors_headless/format_types/juce_LV2PluginFormatImpl.h
+++ b/modules/juce_audio_processors_headless/format_types/juce_LV2PluginFormatImpl.h
@@ -4290,25 +4290,33 @@ public:
     FileSearchPath getDefaultLocationsToSearch()
     {
       #if JUCE_MAC
-        return { "~/Library/Audio/Plug-Ins/LV2;"
-                 "~/.lv2;"
-                 "/usr/local/lib/lv2;"
-                 "/usr/lib/lv2;"
-                 "/Library/Audio/Plug-Ins/LV2;" };
+        return FileSearchPath (SystemStats::getEnvironmentVariable ( "LV2_PATH", { "~/Library/Audio/Plug-Ins/LV2;"
+                                                                                   "~/.lv2;"
+                                                                                   "/Library/Audio/Plug-Ins/LV2;"
+                                                                                   "/usr/local/lib/lv2;"
+                                                                                   "/usr/lib/lv2" })
+                                 .replace (":", ";"));
       #elif JUCE_WINDOWS
-        return { "%APPDATA%\\LV2;"
-                 "%COMMONPROGRAMFILES%\\LV2" };
+        return FileSearchPath (SystemStats::getEnvironmentVariable ("LV2_PATH",
+                                                                    { "%APPDATA%\\LV2;"
+                                                                      "%COMMONPROGRAMFILES%\\LV2" }));
       #else
+        char const* fallback_str = { "~/.lv2;"
+                                     "/usr/local/lib/lv2;"
+                                     "/usr/lib/lv2" };
+
        #if JUCE_64BIT
+        // Check for systems like Fedora (in contrast to e.g. debian)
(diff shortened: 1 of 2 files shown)


hcsch:
As the PR has been sitting around uncommented for a bit I had a quick look at the commit histories for the files I touched in an attempt to reach the right people.
Could I perhaps ask you, @reuk, to take a look at this if you have some time to spare? You seem to be the person who has contributed the most to the LV2 and VST3 plugin format support code in JUCE.

reuk:
Thanks for your patience. We've pushed a similar set of changes here: https://github.com/juce-framework/JUCE/commit/9971c75b2c9d28d8c79d017180bdb3ea58f21158

PR #1626, opened 2026-02-16
Fix invalid note ons on inactive zone's master channel

aggniesz:
MPEZone::isUsing doesn't check if a zone is active so a master channel of an unused zone was being reported as a valid channel, resulting in invalid triggered notes.


Anthony-Nicholls:
> MPEZone::isUsing doesn't check if a zone is active so a master channel of an unused zone was being reported as a valid channel, resulting in invalid triggered notes.

@aggniesz thanks for reporting this should be fixed on the latest develop branch let me know if you experience any more issues.

https://github.com/juce-framework/JUCE/commit/8a14e656f639d0e9f2a630f1d26bd53d3e919d7b

Issue #1334, opened 2024-01-24
[Bug]: Grid gap size rounding issue on Windows


bernidb:
### Detailed steps on how to reproduce the bug

`juce::Grid`'s item bounds calculation with rounding leads to inconsistent behavior on Windows regarding rendering of the gap property. It seems to be independent of the DPI and monitor scaling. This is especially visible if the gap is set to 1px and the grid is resized.
To reproduce:
* open up juce's GridDemo
* set the `grid.rowGap` and `grid.columnGap` to `1_px`
* for better visibility, set the grid item panel colours to `Colours::white`
* launch the demo and observe the behavior while resizing:

![JUCE_Grid_Gap_Bug](https://github.com/juce-framework/JUCE/assets/152602821/3160aa96-09c8-4da2-a6c1-14d0303f9cc6)

Similarly, setting the gap to `2_px` leads to rounding between 1px and 2px and still flickers while resizing.

### What is the expected behaviour?

The expected behavior is that the gap is exactly as wide as set in the variable(s). In the example it should be 1px wide - but it sometimes vanishes due to rounding.

Note: It also seems that the pixel and bounds calculation greatly differs with respect to subpixel handling between `juce::Grid` and `juce::FlexBox`! FlexBox does not have this issue.

### Operating systems

Windows

### What versions of the operating systems?

Tested on Windows 10 x64

### Architectures

64-bit

### Testing on the `develop` branch

The bug is present on the `develop` branch

### Code of Conduct

- [X] I agree to follow the Code of Conduct

PR #1627, opened 2026-02-17
Resolving Grid gap size rounding issue on Windows #1334

Kris-p111:
Changed the rounding function in grid.cpp so that it only applies after the final layout has been computated.


Anthony-Nicholls:
Have you got any examples where the previous behaviour is causing a problem? It would be useful for us to have some test case that we can take a closer look at.

Thanks.

PR #1640, opened 2026-03-28
SystemStats: Make Windows+Clang callCPUID() clear ECX like __cpuid()

why-trv:
This fixes AVX2 and AVX512 detection issues with clang-cl on Windows (https://forum.juce.com/t/bug-report-wrong-systemstats-hasavx2-result-with-clang-cl-64-bit/68580).
Previously, ECX could contain result from a previous callCPUID() call, so the result would be some other leaf, not EAX=7, ECX=0.


tpoole:
Thank you for reporting. I've added a fix here: https://github.com/juce-framework/JUCE/commit/28706b9811e4636e2983c1b686914eb24327d160

PR #1649, opened 2026-04-24
iOS: Reload input view instead of restarting when changing text input target

aggniesz:
Restarting the text input session causes the keyboard to hide and show unnecessarily - no need for that as reloading the input views is enough to update the keyboard type.


reuk:
Thanks! Added here: https://github.com/juce-framework/JUCE/commit/1a7300e7cfb303af29c6559c686aeefbc92a44f0

PR #1657, opened 2026-05-13
Fix builds on non-glibc

WhyNotHugo:
These two patches fix builds on non-glibc Linux (e.g.: musl/Linux):

- Check that `execinfo.h` exists using the `__has_include` before using it. This is more flexible than the current platform-based approach; execinfo will be used on any target where it's present and excluded if absent.
- Use `stat64` workaround only on glibc. Previously this was used unconditionally on any Linux, but this workaround is only necessary on glibc specifically.
- Also check for glibc explicitly for using glibc extensions rather than non-BSD. Same rationale.

### DIFF
--- a/modules/juce_core/juce_core.cpp
+++ b/modules/juce_core/juce_core.cpp
@@ -108,9 +108,10 @@
  #include <net/if.h>
  #include <sys/ioctl.h>
 
- #if ! (JUCE_ANDROID || JUCE_WASM)
-  #include <execinfo.h>
- #endif
+#if defined(__has_include) && __has_include(<execinfo.h>)
+ #define JUCE_HAS_EXECINFO 1
+ #include <execinfo.h>
+#endif
 #endif
 
 #if JUCE_MAC || JUCE_IOS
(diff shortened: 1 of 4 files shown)


WhyNotHugo:
I'm not a lawyer, but I'm pretty sure this CLA has no legal standing.

I was prompted to log in with GitHub, and the immediately presented with a screen that says "_Signing successful. Please download the signed agreement here._". At no point was the agreement itself displayed to be before this screen, nor did I have a choice to agree/refuse.

tpoole:
> At no point was the agreement itself displayed to be before this screen, nor did I have a choice to agree/refuse.

I'm not sure what happened there.

If I open https://cla.juce.com/ then I can see the legal agreement. It's only when you complete the form fields and click on "Submit" that you'll be prompted to log in with GitHub.

WhyNotHugo:
> If I open https://cla.juce.com/ then I can see the legal agreement.

It's only visible _after_ you go through the login process. The initial page doesn't show the agreement, I expected that it would show up in a later step:

image

Tried also with a different browser:

image

I tried checking with `curl https://cla.juce.com/`, and there's no agreement text there, although there's some obfuscated data in a base64 pdf.

Anyway, this is tangential to the submission at hand; I'd rather focus on the fixes.

reuk:
Thanks for reporting, added here:

https://github.com/juce-framework/JUCE/commit/0ae64bf506c4b4e459a2272d2da9fc0c55f9f864

WhyNotHugo:
Thanks!

PR #1658, opened 2026-05-18
Stopped MessageThread quit messages being posted multiple times on MacOS

maxwalley:
On MacOS if multiple quit messages are posted to the message thread before the app has managed to quit an exception of type NSInternalInconsistencyException will be thrown. The easiest way to test this is to create a blank juce GUI app and call juce::JUCEApplicationBase::quit() twice in a row in the initialise() function.

This fix changes it so that if a quit message has been posted, another cannot be posted.

### DIFF
--- a/modules/juce_events/native/juce_MessageManager_mac.mm
+++ b/modules/juce_events/native/juce_MessageManager_mac.mm
@@ -360,6 +360,11 @@ void MessageManager::stopDispatchLoop()
 {
     if (isThisTheMessageThread())
     {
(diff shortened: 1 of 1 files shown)


reuk:
I believe this issue was addressed on the `juce9` branch here: https://github.com/juce-framework/JUCE/commit/eea7d15bba48c9fa9b200f05eb634f16c4a4203d

Anthony-Nicholls:
@maxwalley I've also just cherry-picked the above commit onto the develop branch. It should appear shortly.

Issue #1659, opened 2026-05-18
[Bug]: VST3 hosting re-queries parameter count during parameter initialisation


tjwrona:
## Summary

Some VST3 plug-ins can fail to load while JUCE is initialising the hosted parameter list because
`VST3PluginInstanceHeadless::initialiseParameterList()` re-queries
`IEditController::getParameterCount()` on every loop iteration.

We hit this while loading Neural DSP's `Archetype Gojira X.vst3` on Windows. The plug-in scanned
successfully, but instantiation failed during JUCE's VST3 parameter initialisation path. A local
change that snapshots the parameter count once before the loop resolves the load failure.

## Affected code

In `modules/juce_audio_processors_headless/format_types/juce_VST3PluginFormatImpl.h`,
`initialiseParameterList()` currently does:

```cpp
for (int i = 0; i < editController->getParameterCount(); ++i)
```

Inside that loop, JUCE constructs a `VST3Parameter`, which fetches/caches parameter info via
`getParameterInfo()`. For the affected plug-in, repeatedly querying `getParameterCount()` while
enumerating parameters appears to produce an unstable enumeration boundary.

The nearby helper `getAllParamIDs()` already uses the safer pattern of reading
`getParameterCount()` once before iterating.

## Expected behaviour

JUCE should enumerate the parameter list using a stable count for the current initialisation pass.

## Actual behaviour

JUCE re-queries the VST3 edit controller's parameter count in the loop condition while parameter
wrappers are being constructed, and the affected plug-in fails to load.

## Proposed fix

Related pull request: #1660 

Snapshot the parameter count before the loop:

```cpp
const auto parameterCount = editController->getParameterCount();

for (int i = 0; i < parameterCount; ++i)
```

## Environment

- JUCE branch tested: `develop`
- Platform: Windows
- Plug-in format: VST3
- Affected plug-in observed: Neural DSP `Archetype Gojira X.vst3`

tjwrona:
As an additional test I downloaded the Waveform DAW (which is built on JUCE) and tried loading the same plugin. I noticed similar behavior. The DAW froze for a substantial amount of time before eventually successfully loading the plugin.

The amount of time it froze for was significant (minutes, not seconds)

PR #1660, opened 2026-05-18
VST3: Avoid re-querying parameter count while initialising parameters

tjwrona:
## Summary

This changes the VST3 hosted-parameter initialisation loop to snapshot
`IEditController::getParameterCount()` once before enumerating parameters.

The existing code re-queries `getParameterCount()` in the loop condition while constructing
`VST3Parameter` wrappers and fetching parameter info. We saw Neural DSP's
`Archetype Gojira X.vst3` fail to instantiate through JUCE's VST3 hosting path until this count was
made stable for the initialisation pass.

`getAllParamIDs()` already follows the same snapshot-before-iteration pattern nearby.

Related issue: #1659

## Testing

- Ran `git diff --check`
- Verified the local change fixes loading `Archetype Gojira X.vst3` in the downstream host

### DIFF
--- a/modules/juce_audio_processors_headless/format_types/juce_VST3PluginFormatImpl.h
+++ b/modules/juce_audio_processors_headless/format_types/juce_VST3PluginFormatImpl.h
@@ -3108,7 +3108,9 @@ private:
             cachedParamValues = CachedParamValues { std::move (allIds) };
         }
 
-        for (int i = 0; i < editController->getParameterCount(); ++i)
+        const auto parameterCount = editController->getParameterCount();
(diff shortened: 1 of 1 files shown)


tjwrona:
Note that this patch seemed to also noticeably reduce the plugin load time for other heavy Neural DSP plugins.

reuk:
Thanks for reporting, a fix is available on the juce9 preview branch:

https://github.com/juce-framework/JUCE/commit/f8fa54bec43e2b7c12ae5fc58031080663a18a80

PR #1669, opened 2026-06-09
Fix a rare crash on shutdown due to MidiInput / CriticalSection destruction order

emezeske:
This crash presents as:

> EXCEPTION_ACCESS_VIOLATION_WRITE / 0x24: Fatal Error: EXCEPTION_ACCESS_VIOLATION_WRITE / 0x24
>   ?, in RtlpWaitOnCriticalSection
>   ?, in RtlpEnterCriticalSectionContended
>   ?, in RtlEnterCriticalSection
>   ?, in juce::AudioDeviceManager::CallbackHandler::handleIncomingMidiMessage
>   ?, in `juce::MidiInput::Impl::consume'::`8'::<T>::operator()
> ...
> (20 additional frame(s) were not displayed)

It's due to the order in which `enabledMidiInputs` and `midiCallbackLock` are declared in the header: `midiCallbackLock` is destroyed *before* `enabledMidiInputs`. So the midi input callback can race and access the already-destroyed CriticalSection, causing this crash.

### DIFF
--- a/modules/juce_audio_devices/audio_io/juce_AudioDeviceManager.cpp
+++ b/modules/juce_audio_devices/audio_io/juce_AudioDeviceManager.cpp
@@ -206,6 +206,7 @@ AudioDeviceManager::AudioDeviceManager()
 
 AudioDeviceManager::~AudioDeviceManager()
 {
+    enabledMidiInputs.clear();
     currentAudioDevice.reset();
     defaultMidiOutput.reset();
 }


reuk:
Thanks, I believe this was resolved here: https://github.com/juce-framework/JUCE/commit/2e3e3e92e1a0cf76da33db663abea52045642bd6

emezeske:
thanks!

PR #1672, opened 2026-06-12
Update VST3 module paths on Linux; ./lib64, debian/ubuntu multiarch, /opt

westurner:
- This adds a bunch of paths to vst3 systemPaths list and also filters the list according to filesystem::exists at the time that getModulePaths() is run


westurner (commented):
(Claude Haiku 4.5) suggested the debian/ubuntu multiarch paths. Probably useful for others?

I initially just added /usr/lib64/vst in a patch to the copy of JUCE in BespokeSynth.

reuk:
Please could you test with 8.0.14? That version already uses the `lib64` paths if they exist on a 64-bit system.

I'm closing this since I believe the functionality is already present on the develop branch.

PR #1674, opened 2026-06-18
Fix VST3 COM Ptr incorrect initialization in headless format (#1673)

### LINKED ISSUE #1673: [Bug]: Incorrect Initialisation of VST3 COM Ptr

https://github.com/juce-framework/JUCE/blob/bcfaaf1960487c89263017d441fc0a9bc8b55366/modules/juce_audio_processors_headless/format_types/juce_VST3PluginFormatHeadless.cpp#L77

I think this should be `IncrementRef::no` like the other sites, otherwise the Ptr gets created with a refcount of 2 and leaks when scanning.

This isn't usually spotted as scanning is often done in a separate process or uses the quick module_info.json path.

No leaks.

### Operating systems

macOS, Windows

26.2

### Architectures

Arm64/aarch64

### Stacktrace

```shell

```

VST3

Tracktion Waveform

The bug is present on the `develop` branch

### Code of Conduct

- [x] I agree to follow the Code of Conduct

vlerdf:
### Description
This PR fixes a memory leak during headless VST3 scanning, as reported in #1673.
The VSTComSmartPtr host was incorrectly initialized with IncrementRef::yes, resulting in an incorrect reference count of 2. Changed it to IncrementRef::no to match the standard initialization behavior of other VST3 host context pointers in the codebase.

Fixes #1673.

### DIFF
--- a/modules/juce_audio_processors_headless/format_types/juce_VST3PluginFormatHeadless.cpp
+++ b/modules/juce_audio_processors_headless/format_types/juce_VST3PluginFormatHeadless.cpp
@@ -74,7 +74,7 @@ void VST3PluginFormatHeadless::findAllTypesForFile (OwnedArray<PluginDescription
         if (pluginFactory == nullptr)
             continue;
 
-        VSTComSmartPtr host { new VST3HostContextHeadless(), IncrementRef::yes };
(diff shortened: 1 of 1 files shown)


vlerdf:
This was fixed in develop after this was opened, via e99588c.

PR #1676, opened 2026-06-24
Fix Android API 28 font availability checks

BillyONeal:
Avoid compiling Android 29 font matcher helpers when targeting earlier Android API levels while preserving the existing fallback paths.

This was detected in https://github.com/microsoft/vcpkg/pull/52554

This change was written by GPT 5.5.

### DIFF
--- a/modules/juce_graphics/native/juce_Fonts_android.cpp
+++ b/modules/juce_graphics/native/juce_Fonts_android.cpp
@@ -273,8 +273,12 @@ public:
 
     Typeface::Ptr createSystemFallback (const String& text, const String& language) const override
     {
+       #if __ANDROID_API__ >= 29
         if (__builtin_available (android 29, *))
             return matchWithAFontMatcher (text, language);
+       #else
+        ignoreUnused (text, language);
+       #endif
 
         // The font-fallback API is only available on Android API level 29+
         jassertfalse;
@@ -295,12 +299,15 @@ public:
 
     static Typeface::Ptr findSystemTypeface()
     {
+       #if __ANDROID_API__ >= 29
         if (__builtin_available (android 29, *))
             return findSystemTypefaceWithMatcher();
+       #endif
 
         return from (FontOptions{}.withName ("Roboto"));
     }
 
+   #if __ANDROID_API__ >= 29
     static JUCE_INTRODUCED_IN_29 Typeface::Ptr findGenericTypefaceWithMatcher (const Font& font)
     {
         using AFontMatcherPtr = std::unique_ptr<AFontMatcher, FunctionPointerDestructor<AFontMatcher_destroy>>;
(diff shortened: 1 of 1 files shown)


BillyONeal:
I was expecting there to be a CLA bot or similar to sign but I see now your instructions lead me to a separate PDF form. Checking with legal folks, stand by...

BillyONeal:
OK, CLA signed. Closing and reopening to hopefully kick the bot...

reuk:
Ideally we want to support a minimum API level that's pre-29 while also including the post-29 code-paths in the final binary, so that we get a single binary artefact that can use the nicer font matching behaviour on newer runtimes while degrading gracefully on older runtimes. The proposed approach will disable the new code when the minimum API level is older than 29, which is not what we want.

Perhaps you could try configuring with `-D ANDROID_WEAK_API_DEFS=ON`. That way, the fontmatcher API will be defined using weak symbols, and we can check for the existence of those symbols at runtime.

I tested locally with this CMake configure command:

```
cmake -B cmake-build-android -D CMAKE_SYSTEM_NAME=Android -G Ninja -D CMAKE_TOOLCHAIN_FILE=~/Library/Android/sdk/ndk/30.0.14904198/build/cmake/android.toolchain.cmake -D JUCE_BUILD_EXTRAS=ON -D JUCE_BUILD_EXAMPLES=ON -D ANDROID_PLATFORM=android-24 -D ANDROID_WEAK_API_DEFS=ON
```

When I do this, and then try building a JUCE example, I get a few `-Wunguarded-availability` warnings, that I suspect are caused by the compiler failing to inline unique_ptr destructors in a couple of places. I'll push a fix for those and update this thread once that's done. In the meantime, using an older NDK (28.1.13356709) seems to work fine.

I'm closing this since I believe the correct approach is to enable ANDROID_WEAK_API_DEFS when configuring the build.

BillyONeal:
@reuk Thanks for the review. That didn't seem to be sufficient because even though the new 29 paths are guarded, there are unique_ptr deleters that aren't guarded so clang complains.

The model suggested https://github.com/juce-framework/JUCE/compare/develop...BillyONeal:JUCE:android-api-28-fonts?expand=1 if you're interested in taking a look; I can reopen that as a PR (or you're welcome to use it); I'm not going to do so without your OK though because I don't want to do the "be a middleman between someone who knows what they're doing and an LLM" anti-pattern; just trying to help @toge with adding the new JUCE to vcpkg.

reuk:
> even though the new 29 paths are guarded, there are unique_ptr deleters that aren't guarded so clang complains

Thanks, that's fixed on our develop branch here: https://github.com/juce-framework/JUCE/commit/50c4d6f08fa056abf5ae69105ee18062abd3ad56

Hopefully with that change in place it'll be possible to add the JUCE develop branch to vcpkg.

PR #1677, opened 2026-06-27
Linux: Detect Renoise's out-of-process plug-in host

emezeske:
On Linux, Renoise runs plug-ins in a separate process named AudioPluginServer_*, so the host filename never contains "Renoise" and PluginHostType could not identify the host. The versioned install path (.../renoise-X.Y.Z/...) does contain it, so test the host path as well. This mirrors the Windows and macOS branches, which already fall back to the host path to recognise several hosts.


szarvas:
Thank you for reporting this. A fix has been released
https://github.com/juce-framework/JUCE/commit/99ba6a4656e3a5b5eb27adc71a3c5feee42d5120

emezeske:
Thanks!

PR #1689, opened 2026-07-22
Audio formats: Harden WAV/AIFF parsers against three malformed-input issues

githappens:
Three memory-safety / DoS fixes in the WAV and AIFF parsers, reached through
the normal createReaderFor + read path on an untrusted file. Each is a
separate commit.

### 1. Clamp metadata chunk length before allocating
The `bext`/`smpl`/`inst`/`cue` WAV chunks and the AIFF `INST` chunk pass an
untrusted 32-bit chunk length straight to `HeapBlock::calloc` before checking
the stream holds that many bytes. A tiny file declaring a ~4 GB chunk requests
a ~4 GB allocation: benign lazy zero-fill on 64-bit, but a failed allocation
and null dereference on 32-bit. Clamp the length to `getNumBytesRemaining()`
first.

### 2. Reject a sample frame larger than the read buffer
`readSamples` copies whole frames through a fixed 5760-byte temp buffer, sizing
each batch with `tempBufSize / bytesPerFrame`. `bytesPerFrame` comes from
untrusted header fields with no upper bound, so a frame larger than the buffer
truncates that division to zero: no samples are consumed and the loop never
terminates — a tiny malformed file hangs the reading thread on every platform.
Bail out before the loop when a single frame can't fit. Affects both the AIFF
and WAV readers (the WAV `bitsPerSample <= 32` gate doesn't prevent it — a high
channel count reaches the same state).

### 3. Bounds-check the cate tag read
`CATEChunk::read` walks the chunk body with a `while (data < dataEnd)` guard
that only guarantees one byte is in bounds, but `isValidTag` dereferences three.
When `data` lands one or two bytes short of the end, it reads past the block.
Require three bytes of headroom, matching the guards already used later in the
same loop.

### DIFF
--- a/modules/juce_audio_formats/codecs/juce_AiffAudioFormat.cpp
+++ b/modules/juce_audio_formats/codecs/juce_AiffAudioFormat.cpp
@@ -549,6 +549,7 @@ public:
                     }
                     else if (type == chunkName ("INST"))
                     {
+                        length = (uint32) jmin ((int64) length, input->getNumBytesRemaining());
                         HeapBlock<InstChunk> inst;
                         inst.calloc (jmax ((size_t) length + 1, sizeof (InstChunk)), 1);
                         input->read (inst, (int) length);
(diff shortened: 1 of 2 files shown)


reuk:
Thanks! Added here:

https://github.com/juce-framework/JUCE/commit/89edd1811fa9526653c54207887e86eaf44eadaa

https://github.com/juce-framework/JUCE/commit/6e65c729523cbebf5715b622c2513b0ac8d63190

PR #1694, opened 2026-08-03
Fix: prevent duplicate mousemove listeners in WebViewPluginDemoGUI

alex-js-ltd:
## Summary
- `App` previously registered a `document` `mousemove` listener directly in the component body (not inside `useEffect`), so a new listener was added on every re-render without the old one being removed — a listener leak.
- Moves the listener registration into a `useEffect` with an empty dependency array and adds a cleanup function that removes it on unmount.

Discussed on the forum: https://forum.juce.com/t/event-listener-leak-in-the-webviewplugindemogui/69236

## Test plan
- [x] Run the WebView plugin demo GUI and confirm mouse-move driven parameter control still works.
- [x] Verify only one `mousemove` listener is attached after multiple re-renders (e.g. via `getEventListeners(document)` in devtools).

### DIFF
--- a/examples/Plugins/WebViewPluginDemoGUI/src/App.tsx
+++ b/examples/Plugins/WebViewPluginDemoGUI/src/App.tsx
@@ -76,7 +76,7 @@ function JuceSlider({
 
   const changeCommitted = (
     _event: SyntheticEvent | Event,
-    newValue: number | number[]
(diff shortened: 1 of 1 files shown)


szarvas:
Thank you for reporting. The change has been merged.

PR #1719, opened 2026-08-27
WASAPI: Fix uninitialised session GUID passed to IAudioClient::Initialize

### LINKED ISSUE #1718: [Bug]: WASAPI: uninitialized session GUID passed to IAudioClient::Initialize splits one app across Windows audio sessions

`WASAPIDeviceBase::initialiseStandardClient` (`modules/juce_audio_devices/native/juce_WASAPI_windows.cpp`, line 888 in 9.0.1) passes an uninitialized stack `GUID` as the `AudioSessionGuid` argument of `IAudioClient::Initialize`:

```cpp
GUID session;
auto hr = client->Initialize (isExclusiveMode (deviceMode) ? AUDCLNT_SHAREMODE_EXCLUSIVE
                                                           : AUDCLNT_SHAREMODE_SHARED,
                              getStreamFlags(),
                              defaultPeriod,
                              isExclusiveMode (deviceMode) ? defaultPeriod : 0,
                              (WAVEFORMATEX*) &format,
                              &session);
```

`AudioSessionGuid` is an input-only parameter (`LPCGUID`), so every standard-mode stream joins an audio session named by whatever the stack holds. Audio sessions are the rows of the Windows Volume Mixer and the key Windows persists per-application volume under, so the garbage is user-visible.

Minimal reproduction (single-file console app that opens devices through `AudioDeviceManager` and prints this process's audio sessions via `IAudioSessionManager2` after each step, observed output included): https://github.com/jdf/juce-wasapi-session-repro

What it shows on Windows 11 with JUCE 9.0.1:

1. The standard-mode stream's session GUID comes out as `{00200016-0003-0000-0300-000000001000}` — not `GUID_NULL`, and not a GUID anything chose. Decoded as bytes it is a stale fragment of the `WAVEFORMATEXTENSIBLE` staged earlier in the same open: `cbSize` 22, `wValidBitsPerSample` 32, `dwChannelMask` 3, then the first eight bytes of `KSDATAFORMAT_SUBTYPE_IEEE_FLOAT`.
2. The leftovers are deterministic for one configuration and call path, so a tight reopen loop hides the bug — but two paths whose leftovers differ split the application across sessions. Switching the device type to "Windows Audio (Low Latency Mode)" and back leaves the process with two sessions on one endpoint, because `initialiseLowLatencyClient` correctly passes `nullptr` (the default session) while the standard path passes garbage. Both rows outlive their streams.
3. Control: with `&session` replaced by `nullptr` and nothing else changed, every phase of the program shares the endpoint's default session — one Volume Mixer row.

This appears to be the mechanism behind a long-standing forum report of `setCurrentAudioDeviceType` creating multiple app entries in the mixer (https://forum.juce.com/t/setcurrentaudiodevicetype-creates-multiple-instances-of-standalone-app-within-windows-mixer-console/31282), and end users of our JUCE application report the same: duplicate rows in the Windows 11 Volume Mixer, one inert, with more appearing after output-device changes.

A related nit in the same code path: JUCE never calls `SetDisplayName` on the `IAudioSessionControl` it acquires, so Windows labels each new row from whatever window is foreground when the session first appears — users see mixer rows titled after settings dialogs.

Shared-mode streams join the endpoint's default audio session (`AudioSessionGuid = nullptr`), as the low-latency path already does: one stable Volume Mixer row per application, whose volume setting persists across device reopens.

### Operating systems

Windows

Windows 11

### Architectures

x86_64

Standalone

The reproduction ran against 9.0.1; the line is byte-identical on `develop` as of 2026-08-27 (`GUID session;` at line 888).

jdf:
Fixes #1718.

`WASAPIDeviceBase::initialiseStandardClient` passes an uninitialized
stack `GUID` as the `AudioSessionGuid` argument of
`IAudioClient::Initialize`. The parameter is input-only (`LPCGUID`), so
every standard-mode stream joins an audio session named by whatever the
stack holds — in practice a stale fragment of the
`WAVEFORMATEXTENSIBLE` staged earlier in the same open. Audio sessions
are the rows of the Windows Volume Mixer and the key Windows persists
per-application volume under, so the garbage is user-visible: streams
that reach `Initialize` through different code paths land in different
sessions, and the application accumulates mixer rows whose volume
settings don't carry over (for example when toggling between the
standard and low-latency device types, or driving an
`AudioDeviceSelectorComponent` — the symptom reported in
https://forum.juce.com/t/setcurrentaudiodevicetype-creates-multiple-instances-of-standalone-app-within-windows-mixer-console/31282).

The fix passes `nullptr`, joining the endpoint's default session, as
`initialiseLowLatencyClient` already does.

The unit test opens the default output through `AudioDeviceManager`
("Windows Audio" shared mode) and asserts that every audio session
belonging to the process — enumerated via `IAudioSessionManager2`
across active render endpoints — is the default session (`GUID_NULL`
in its instance identifier). It skips on machines with no output
device. Verified on Windows 11 x64 (MSVC, Ninja, `UnitTestRunner
--name "WASAPI audio sessions"`): the test fails against the current
code with `a stream joined a non-default audio session:
…%b{00200016-0003-0000-0300-000000001000}|…` and passes with the fix.

A standalone reproduction with observed output, including the
byte-decode showing the garbage GUID is leftover format data, is at
https://github.com/jdf/juce-wasapi-session-repro.

### DIFF
--- a/modules/juce_audio_devices/native/juce_WASAPI_windows.cpp
+++ b/modules/juce_audio_devices/native/juce_WASAPI_windows.cpp
@@ -885,14 +885,15 @@ private:
 
         for (;;)
         {
-            GUID session;
+            // A null session GUID joins the endpoint's default audio session,
+            // so the application keeps a single Volume Mixer entry.
             auto hr = client->Initialize (isExclusiveMode (deviceMode) ? AUDCLNT_SHAREMODE_EXCLUSIVE
                                                                        : AUDCLNT_SHAREMODE_SHARED,
                                           getStreamFlags(),
                                           defaultPeriod,
                                           isExclusiveMode (deviceMode) ? defaultPeriod : 0,
                                           (WAVEFORMATEX*) &format,
-                                          &session);
+                                          nullptr);
 
             if (check (hr))
                 return true;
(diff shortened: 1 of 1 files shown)


jdf:
I have now signed the CLA.

szarvas:
Merged in https://github.com/juce-framework/JUCE/commit/b023b56f2b0e6c68e76fec19ce68a0f918af18b0

PR #1731, opened 2026-09-06
Linux: Open GTK and WebKit by their versioned library names as well

### LINKED ISSUE #1730: Linux: WebBrowserComponent shows nothing unless the GTK and WebKit development packages are installed

1. On Ubuntu 24.04 or 22.04 (Debian and Fedora behave the same), install the runtime packages only: `libwebkit2gtk-4.1-0` (or `libwebkit2gtk-4.0-37`) and `libgtk-3-0`, and make sure no `-dev` package for them is installed.
2. Run any JUCE application or plugin that shows a `WebBrowserComponent` and call `goToURL()` on it, for example the DemoRunner's WebBrowserDemo built elsewhere and copied over.

The page renders.

### What is the actual behaviour?

The component stays blank and nothing is reported. `WebKitSymbols` in `modules/juce_gui_extra/native/juce_WebBrowserComponent_linux.cpp` opens the libraries with `dlopen` under their unversioned names (`libgtk-3.so`, `libglib-2.0.so`, `libwebkit2gtk-4.1.so`, `libjavascriptcoregtk-4.1.so`, `libsoup-3.0.so`, and the 4.0 equivalents). Distributions ship those unversioned names only in the development packages, as a symlink to the versioned library; a system with just the runtime packages has `libwebkit2gtk-4.1.so.0` and `libgtk-3.so.0`. Every `dlopen` fails, `isWebKitAvailable()` is false, and `Platform::init()` returns before launching the child process, so `goToURL` does nothing.

On a user's machine (Ubuntu, no development packages):

```
$ ls -l /usr/lib/x86_64-linux-gnu/libwebkit2gtk-4.1.so* /usr/lib/x86_64-linux-gnu/libgtk-3.so*
lrwxrwxrwx 1 root root       21 Jan 27  2025 /usr/lib/x86_64-linux-gnu/libgtk-3.so.0 -> libgtk-3.so.0.2409.32
-rw-r--r-- 1 root root  8156632 Jan 27  2025 /usr/lib/x86_64-linux-gnu/libgtk-3.so.0.2409.32
lrwxrwxrwx 1 root root       28 Aug 21 19:04 /usr/lib/x86_64-linux-gnu/libwebkit2gtk-4.1.so.0 -> libwebkit2gtk-4.1.so.0.21.10
-rw-r--r-- 1 root root 94901008 Aug 21 19:04 /usr/lib/x86_64-linux-gnu/libwebkit2gtk-4.1.so.0.21.10
```

Reproduced in a container built from Ubuntu 22.04 with the development packages installed: the view renders; after removing only the unversioned symlinks (`libgtk-3.so`, `libglib-2.0.so`, `libwebkit2gtk-4.0.so`, `libjavascriptcoregtk-4.0.so`, `libsoup-2.4.so`), the same binary shows a blank view.

Trying each library under its versioned name as a fallback (`libwebkit2gtk-4.1.so.0`, `libjavascriptcoregtk-4.1.so.0`, `libsoup-3.0.so.0`, `libgtk-3.so.0`, `libglib-2.0.so.0`, and `libwebkit2gtk-4.0.so.37`, `libjavascriptcoregtk-4.0.so.18`, `libsoup-2.4.so.1`) makes the view render on that system. Pull request to follow.

### Operating systems

Linux

Ubuntu 24.04 (reported), Ubuntu 22.04 (reproduced)

### Architectures

x86_64

### Stacktrace

None; nothing crashes.

VST3 and a standalone application, identically.

_No response_

The bug is present on the `develop` branch (04e167d)

### Code of Conduct

jdf:
Fixes #1730.

`WebKitSymbols` opens GTK, GLib, WebKit, JavaScriptCore, and libsoup with `dlopen` under their unversioned names, which distributions install only with the development packages. A system with the runtime packages alone has `libwebkit2gtk-4.1.so.0` and `libgtk-3.so.0`, every open fails, and `WebBrowserComponent` shows nothing.

Each library is now tried under its unversioned name and then its versioned one: `.so.0` for GTK, GLib, and the 4.1 set; `.so.37`, `.so.18`, and `.so.1` for WebKit, JavaScriptCore, and libsoup in the 4.0 set. `DylibHandle` gains a constructor taking the list of names and keeping the first that opens.

Verified in a container built from Ubuntu 22.04: with the unversioned symlinks removed, an application's web view showed nothing before the change and renders a page after it; with the symlinks present it renders both before and after.

### DIFF
--- a/modules/juce_gui_extra/native/juce_WebBrowserComponent_linux.cpp
+++ b/modules/juce_gui_extra/native/juce_WebBrowserComponent_linux.cpp
@@ -280,6 +280,21 @@ private:
         DylibHandle (const char* str, int flags)
             : handle (dlopen (str, flags)) {}
 
+        /*  The first of the names that opens. A distribution ships the
+            versioned name with the library and the unversioned one only
+            with the library's development package, so a system that has
+            never built against GTK or WebKit carries only the former.
+        */
+        explicit DylibHandle (std::initializer_list<const char*> names)
+            : DylibHandle (names, RTLD_NOW | RTLD_LOCAL) {}
+
+        DylibHandle (std::initializer_list<const char*> names, int flags)
+        {
+            for (auto* name : names)
+                if ((handle = dlopen (name, flags)) != nullptr)
+                    break;
+        }
+
         ~DylibHandle()
         {
             if (handle != nullptr)
@@ -434,9 +449,9 @@ private:
 
     struct WebKitAndDependencyLibraryNames
     {
-        const char* webkitLib;
-        const char* jsLib;
-        const char* soupLib;
+        std::initializer_list<const char*> webkitLib;
+        std::initializer_list<const char*> jsLib;
+        std::initializer_list<const char*> soupLib;
     };
 
     bool openWebKitAndDependencyLibraries (const WebKitAndDependencyLibraryNames& names)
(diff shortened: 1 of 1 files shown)


szarvas:
Thank you for reporting. A fix is out on develop
https://github.com/juce-framework/JUCE/commit/146fdea555ee25c3a9e9a41feeeb4e72919cccb6

PR #1744, opened 2026-09-11
Fix file unit test under elevated permissions

### LINKED ISSUE #1724: [Bug]: UnitTestRunner fails when run as root

~~~command
$ sudo ./UnitTestRunner --categories=Files

Random seed: 0xc7693885cd76f736
-----------------------------------------------------------------
Starting tests in: Files / Reading...
Completed tests in Files / Reading
-----------------------------------------------------------------
Starting tests in: Files / Writing...
!!! Test 47 failed
FAILED!!  1 test failed, out of a total of 57

-----------------------------------------------------------------
Starting tests in: Files / Memory-mapped files...
Completed tests in Files / Memory-mapped files
-----------------------------------------------------------------
Starting tests in: Files / More writing...
Completed tests in Files / More writing
-----------------------------------------------------------------
Starting tests in: FileSearchPath / removeRedundantPaths...
Completed tests in FileSearchPath / removeRedundantPaths

-----------------------------------------------------------------
Test failure summary:

Files / Writing: 1 test failure
	!!! Test 47 failed

~~~

i should be able to run the UnitTestRunner as any user.

if a test makes certain assumptions (like: "the user does not have superpowers which allows them to write a read-only file"), this test should be skipped if the assumptions are not met.

while usually you probably shouldn't run unit tests as root (because they might break your system), i find this a common setup in CI system (docker, VMs,... should isolate your actual system good enough)

### Operating systems

Linux

Debian

### Architectures

x86_64, 32 bit Arm (Linux, Android)

### Stacktrace

```shell

```

_No response_

_No response_

I have not tested against the `develop` branch

### Code of Conduct

- [x] I agree to follow the Code of Conduct

iibaranov-IG:
Running the file unit tests as root (a common Docker/CI setup) currently fails in `Files / Writing`: `File::hasWriteAccess()` deliberately returns true for an effective root user, so the test's unconditional read-only access assertion cannot hold.

This change still verifies that both `setReadOnly(true)` and `setReadOnly(false)` succeed. It only checks the observable write-access transition when the environment enforces it, and logs an explicit skip when elevated permissions bypass the restriction.

Validation on Ubuntu 24.04 in a root Docker container:

- Before: `Files / Writing: 1 test failure`, `Test 47 failed`
- After: `./UnitTestRunner --category=Files` completes all five file subtests successfully and logs the elevated-permission skip

Fixes #1724


tpoole:
Thank you for reporting. Fixed here: https://github.com/juce-framework/JUCE/commit/344cdbc112c0a8ef60da5e0e3f4845d61ea3fdd8

